Issue #20, opened 2021-08-13
Burrito does not create a shared memory file


AsherGlick:
In order to access the Mumble Link API a chunk of memory must be allocated. Burrito should do this automatically but does not currently. Early testing worked because Arcdps creates the shared memory file though so gw2 starts writing to it when that happens. But if you launch GW2 without arcdps then Burrito will crash because it is trying to read from an out of bounds memory address because nothing has been linked to it.

AsherGlick:
No I think #26 fixed this

PR #26, opened 2021-08-15
creates shared memory if it doesn't exist rather than depending on arcdps creating it

coderedart:
just copy pasted from taco's implementation.

### DIFF
--- a/Spatial.tscn
+++ b/Spatial.tscn
@@ -17,10 +17,10 @@ void fragment(){
 [sub_resource type="ShaderMaterial" id=2]
 shader = SubResource( 1 )
 
-[sub_resource type="PlaneMesh" id=4]
+[sub_resource type="PlaneMesh" id=3]
 size = Vector2( 76.2, 76.2 )
 
-[sub_resource type="ShaderMaterial" id=5]
+[sub_resource type="ShaderMaterial" id=4]
 shader = ExtResource( 2 )
 shader_param/line_color = null
 shader_param/line_width = 0.044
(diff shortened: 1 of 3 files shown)


coderedart:
closes https://github.com/AsherGlick/Burrito/issues/20

AsherGlick:
This is awkward and I apologize but I cannot review this PR. The taco dev has been very adamant about what their software license is, and it is incompatible with the GPL.

If this PR is just a couple lines that does something like
```
if (size of(memory file) == 0) {
    create_memory_file(MAX_MEMORY_FILE_SIZE)
}
```
then I can merge it if you can find the example to it in public documentation and use theirs instead. But I cannot pull code copied from taco without the original author's approval. (see taco issue number 2)

coderedart:
It's just a syscall. Nothing about it is specific to taco.  just mentioned it as a futures reference to the related code from similar project. And this is c code which taco is CPP code, that's why we need to use sizeof (struct LinkedMem) compared to taco.
  More references
https://docs.microsoft.com/en-us/windows/win32/memory/creating-named-shared-memory

https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createfilemappinga

coderedart:
But if you insist I will make a different pr tomorrow 😜

coderedart:
kinda my first PR. so i messed it up. help me Asher sempai :) i added 3 files instead of just the burrito_link file

AsherGlick:
I recommend using [git restore](https://git-scm.com/docs/git-restore) to roll back the files you did not intend to submit and make a new commit that reverts them. After submitting you can use git restore again to roll back the roll back locally so you still have your changes on your machine.

coderedart:
it feels big because of formatting changes, but i tried my best to stick to the reference docs of microsoft. checked that it works.

AsherGlick (approved):

Issue #11, opened 2021-08-13
Binary Burrito Files


AsherGlick:
Burrito files are currently stored in `.json` format. Most of the data in them is either structural data `{}``[]` or floating point numbers. These are traditionally poorly encoded in the JSON format.

Using protocol buffers is a good compromise between a strict custom packed binary file and JSON formatting.

Another option instead of protocol buffers is to use Godot's binary serialization. The upside to this over protocol buffers is that it is built into the godot codebase and additional protobuf code will not need to be created. A downside to this is that it is not schema based meaning that it would be harder to handle future proofing the file. Using the built-in serialize would also make it more difficult for other programs to interact with it. This may still be an option for a short-term solution.

coderedart:
I definitely think markers format should move to binary format as with burrito, it's not necessary to be manually editing XML files anymore. 


1. Would be more useful to decide with other overlays developers.
2. Must be a format with multiple languages support like protobuffers or flat buffers or even flex buffers. 

Although we will still need to provide a XML exporter or converter if we want people to use burrito to create markers. People definitely want compatibility with taco if they can get it

AsherGlick:
Windows support is tracked in #9.

dlamkins:
> Although we will still need to provide a XML exporter or converter if we want people to use burrito to create markers. People definitely want compatibility with taco if they can get it

This is an incredibly important point.  Please sure to work with other overlays to help maintain compatibility in one way or another.  We want to continue to give marker pack makers powerful tools and features without splintering their userbases.

Blish HUD can handle alternative format readers (our reader is handled by a library called `TmfLib` we've written to keep the spec mostly straightforward), but we have had no interest in introducing other formats at the moment because we've not yet found any reasonable benefit to it.

We (Blish HUD) keeps an [attribute implementation list](https://blishhud.com/docs/markers/attributes/achievement) if you wish you match the implementations of TacO & Blish HUD and we can note your compatibility there.

AsherGlick:
Interoperability is definitely a good thing. Right now all there is is [this script](/AsherGlick/Burrito/blob/0e95e21e0fe1f2eec36aa0901b3197dc7ff4153f/scripts/taco_to_burrito.py) to convert from XML to the Burrito format, I plan to update it when a binary file format is implemented. A similar converter could be created to convert back.

However, the set of data stored in these formats would possibly be disjoint depending on the outcome of these issues:

[...]

Issue #3, opened 2021-08-13
Marker Navigation Menu


AsherGlick:
It is unclear what the best design for implementing a menu where markers can be enabled or disabled. At the time of opening Burrito only opens files and displays their entire contents. However there are situations where you want multiple marker files to be displayed at the same time. Such as a combination of Tekkit Workshop's map completion markers combined with an overlay of the boundaries of all the Task Hearts.

If enabling / disabling subsets of markers is deemed to be not-the-best use case then it should also be decided if a file UI is the best interface or if a custom UI would be better.

PR #28, opened 2021-08-15
scripts: improve taco_to_burrito.py

TarEnethil:
slight improvements for taco_to_burrito.py I made while using it to
convert Teh's Trails.

- add rudimentary support for command line arguments (src + dest). note:
this is not good python code, but as I understand it, converting via
python is only supposed to be a temporary solution.
- also create DEST_DIR/data directory to prevent an error-on-first-write
- support Trails-as-MarkerCategory by adding the texture attribute if it
is present in the <MarkerCategory> and using it when generating Trail
data

with these changes, I was able to convert Teh's Trails into a working
burrito format, after making some adjustments to the original files
(fixing some file names and removing all but one color of trails in the
xml).

Signed-off-by: Thorben Römer <thorbenroemer@t-online.de>

### DIFF
--- a/scripts/taco_to_burrito.py
+++ b/scripts/taco_to_burrito.py
@@ -5,15 +5,26 @@ import struct
 import shutil
 from typing import *
 import json
+import sys
 
 output_dir = ""
 
-source_path: str = "tw_ALL_IN_ONE"
-target_path: str = "tekkit_workshop_burrito_markers"
+source_path: str = ""
+target_path: str = ""
 
 def main():
+    if len(sys.argv) != 3:
+        print(f"USAGE: {sys.argv[0]} SRC_DIR DEST_DIR")
+        sys.exit(1)
+
+    global source_path
+    global target_path
+
+    source_path = sys.argv[1]
+    target_path = sys.argv[2]
+
     if not os.path.exists(target_path):
-        os.makedirs(target_path)
+        os.makedirs(os.path.join(target_path, "data"))
 
     for file in os.scandir(source_path):
         if(file.name.endswith(".xml")):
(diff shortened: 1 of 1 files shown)


scripts/taco_to_burrito.py
```diff
@@ -90,6 +101,9 @@ def parse_marker_category(marker_category_node, base=""):
     else:
         limited_attribs["height_offset"] = 0
 
+    if "texture" in attribs:
```
AsherGlick, line 104: Does `"texture"` need to be added to the `allowed_keys` above? I feel like this would cause an error if the texture attribute existed in the marker category

AsherGlick (commented):
You are correct that this script is ideally only temporary. Is https://github.com/xrandox/TehsTrails/ the trail file you are converting?

scripts/taco_to_burrito.py
TarEnethil: Technically speaking yes, although "cause an error" in this case only means "print an error message". I've added all previously unknown attributes that I've encountered while converting Teh's Trails.

I'm not really sure why most of those checks are done, it seems most 'required' attributes are never used anyway.

TarEnethil: I've indeed used Teh's Trails from the link you posted. Although some filenames and references are broken ([PR opened](https://github.com/xrandox/TehsTrails/pull/20)). The 'only' remaining issue is that multiple identical Trails (different texture/colors) are defined, which overlap and clip through one another. I think in TacO you can just select / put checkboxes on the Trails you want to see. For me personally I removed all but one of the Trail definitions.

AsherGlick: Good point about the required attributes not being used. The goal was reverse engineering the spec not dictating it. So if it looked like an attribute was required in the spec it would go into "required" until a counterexample was found, then it would be moved to allowed. If an error was printed about a key not being allowed that just meant that some attribute had not yet had human consideration for how it should be converted.

I think you are correct about TacO's checkboxes BlishHUD definitely does that and there is some conversation on it in #11 and #3 .

Issue #25, opened 2021-08-14
Burrito loses focus when focusing on Gw2


Nuxmin:
Hey there!

I could make it work like you stated using ArcDPS, however, another issue I've seen. Even if the Burrito is checked as always on top and I have gw on windowed. Every time you focus on gw the Burrito stays in the background.

I find no way to make it stay on top so it's only possible to see the path alternating with Alt Tab.

AsherGlick:
Can you share what distro and window manager you are using?

Nuxmin:
Sure!

I'm using Pop OS, it uses GNOME.

AsherGlick:
Is guild wars in fullscreen or windowed-fullscreen?

Nuxmin:
> Is guild wars in fullscreen or windowed-fullscreen?

Windowed.

AsherGlick:
Unfortunately I cannot reproduce this issue on ubuntu using gnome and I do not have the resources to replicate a popos environment. If you or anyone else happen to figure it out, or are able to debug any more into the issue please add it here so we can work on a fix.

coderedart:
i have the same issue. Endeavour/KDE latest packages. 

so, if you click on gw2 window, the whole overlay goes into background. but if you set the gw2 window as "keep below others" , overlay will now display over gw2 window. 
the issue with this work around is that you can't focus on burrito by clicking on its icon, you have to use alt-tab to actually focus input on burrito and then, you can click on burrito icon to get the main panel of "open marker file" and such. 

i just recommend sticking to window mode for now and maybe play gw2 with "borderless" decoration.

coderedart:
i checked on gnome 40 and xfce 4.16 . the same issue with windowed full screen. with gnome, i couldn't even use windowed mode of gw2. 

According to distrowatch,
pop os has gnome 40.2 or 40.3, while arch/endeavour got gnome 40.3 . 
OTOH, ubuntu has the following gnome versions with the respective ubuntu releases. 
3.38.4 | 3.38.1 | 3.36.1
21.04  | 20.10  | 20.04 lts.

I *think* gnome 4 might have a bug that affects me and @Nuxmin . will need more people to report their distros/DE versions along with what issues they have with burrito to confirm. @AsherGlick what's your distro/gnome version? maybe you can put it in README as the main target that you develop for?

coderedart:
https://github.com/AsherGlick/Burrito/issues/23#issuecomment-899065487
so, cinnamon works i guess.

[...]

PR #32, opened 2021-08-31
Added gdnative script to set X11 transient_for property

royalmustard:
This pr adds a gdnative library (written in rust) to solve #25.
The library exposes a method which takes Burritos X Window id from godot.
It then acquires the GW2 window Id by searching for a window with the name "Guild Wars 2". Because of that burrito needs to be launched after GW2 is already running.
Finally, it sets the WM_TRANSIENT_FOR property of the Burrito window.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 .import/
 build/*
+burrito-fg/target/*
+burrito-fg/Cargo.lock
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 .import/
 build/*
 burrito-fg/target/*
-burrito-fg/Cargo.lock
--- a/burrito-fg/Cargo.toml
+++ b/burrito-fg/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "burrito-fg"
+version = "0.1.0"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+gdnative = "0.9"
+breadx = {git="https://github.com/royalmustard/breadx"}
--- a/burrito-fg/src/lib.rs
+++ b/burrito-fg/src/lib.rs
@@ -1,0 +1,91 @@
+use breadx::{
+    auto::xproto::ATOM_WM_TRANSIENT_FOR,
+    traits::{PropMode, PropertyFormat, PropertyType, ATOM_WM_NAME},
+    Display, DisplayBase, DisplayConnection, Window,
+};
+use gdnative::prelude::*;
+
+#[derive(NativeClass)]
+#[inherit(Node)]
+pub struct X11_FG;
+// Function that registers all exposed classes to Godot
+fn init(handle: InitHandle) {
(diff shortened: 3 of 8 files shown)


.gitignore
```diff
@@ -1,2 +1,4 @@
 .import/
-build/*
\ No newline at end of file
+build/*
+burrito-fg/target/*
+burrito-fg/Cargo.lock
```
AsherGlick, line 4: We should be including the Cargo.lock file in git. It contains helpful information like the package checksum and exact package version.

burrito-fg/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+gdnative = "0.9"
+breadx = {git="https://github.com/royalmustard/breadx"}
```
AsherGlick, line 12: Is there a reason we are not using the mainline breadx crate?

burrito-fg/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+#[methods]
+impl X11_FG {
+    #[export]
+    fn _ready(&self, _owner: &Node) {
+        godot_print!("Hello,X11!");
+        return;
+    }
+
+    #[export]
+    fn set_transient_for(&self, _owner: &Node, burrito_id: i32) {
```
AsherGlick, line 32: Can we add a comment here as to why we are taking in an `i32` only to immediately cast it to a `u32`. I'm guessing it has something to do with the gdnative api but it would be nice to have a clear record here.

AsherGlick (commented):
This is great!

As far as needing to launch burrito after gw2, I think that is fine. If it becomes something that is higher priority in the future we can fix it then. Just having it work properly for people affected by this issue in the first place is a giant win.

This PR also seems to have broken the build system. This is because there is nothing building the new rust library inside of [this workflow](https://github.com/AsherGlick/Burrito/blob/master/.github/workflows/main.yml). I think it would be best to include the build commands in that workflow, hopefully just a `cargo build` is required.

burrito-fg/Cargo.toml
royalmustard: Mainline breadx has a bug causing it to panic when it deserializes an empty string property from X. I already mad a PR, but for now we are stuck with the fix

burrito-fg/src/lib.rs
royalmustard: Its because the gdnative api allows only types from godot to be given to a method.

burrito-fg/Cargo.toml
AsherGlick: SGTM, can you add a new issue once this is merged to update this dependency whenever your pr gets merged into breadx?

burrito-fg/src/lib.rs
AsherGlick: Makes sense to me.

royalmustard:
Breadx has merged my PR: Should we change it to the mainline git repo or wait until the next crates.io release of breadx?

PR #34, opened 2021-09-14
Change breadx dependency to crates.io

royalmustard:
as my changes have been merged and the new diagnostics for connecting were added, there is no more reason to use my repo for this


AsherGlick:
Great! glad to hear your changes were published.

PR #36, opened 2021-09-23
Use xid supplied by link to set transient

### LINKED ISSUE #33: Burrito_FG fails to connect to X on some configurations
as discussed in #25 the current implementation of burrito-fg fails to connect to the x server sometimes. This seems to be an issue with breadx.

royalmustard:
Should fix #33 for now. We get the xid of gw2 from the link. 
Inside the decode_context_packet method we try to set the transient_for property, if it has not already been set.


AsherGlick (changes requested):
This looks great.

As for the extra file changes: Godot sometimes makes some automatic readjustments. In general they are fine but there are a short series here you should manually revert because they contain info about your local system that is not required inside of the repo.

AsherGlick (approved):

PR #38, opened 2021-09-27
Add exit button to menu

royalmustard:
This was bugging me for quite a while.
Since Godot seems to overwrite the file paths every time, maybe we should consider just leaving them in or come up with another solution.

### DIFF
--- a/Spatial.gd
+++ b/Spatial.gd
@@ -745,3 +745,7 @@ func _on_SetActivePath_pressed():
 
 func _on_ReversePathDirection_pressed():
 	self.currently_selected_node.object_link.reverse()
+
+
+func _on_ExitButton_pressed():
+	get_tree().quit()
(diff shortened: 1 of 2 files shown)


AsherGlick:
I don't think it is a huge lift to remove them manually prior to committing. Maybe there is a presubmit script that can be run to take care of this. Given how simple the `.tscn` files are to parse it could just be a simple regex replace for an entire line.

```bash
sed --in-place -e 's/current_dir = "[^"]*"/current_dir = "."/g' *.tscn
sed --in-place -e 's/current_file = "[^"]*"/current_file = "."/g' *.tscn
sed --in-place -e 's/current_path = "[^"]*"/current_path = "."/g' *.tscn
```

There seems to be an open bug about this too https://github.com/godotengine/godot/issues/29674

PR #41, opened 2021-10-08
Taco xml parser

jlloh:
Wrote a parser in rust to parse taco xml files and the `trl` c-struct [here](https://github.com/jlloh/gw2_taco_parser). 

The helper util will return a json string that's backward compatible such that Burrito's existing `gen_map_markers()` can be used. This util is used if an xml extension is detected. 

At the moment, it still assumes that the `.taco` file is "unzipped" and in a folder.

Let me know what you guys think of this approach. I placed the `gw2_taco_parser` in it's own crate so to keep the code cleaner.

### DIFF
--- a/taco_parser/src/xml_parser.rs
+++ b/taco_parser/src/xml_parser.rs
@@ -99,9 +99,8 @@ struct Icon {
 }
 
 //https://docs.rs/serde-xml-rs/0.5.1/serde_xml_rs/
-pub fn parse_xml(xml: &str) -> OverlayData {
-    let data: OverlayData = serde_xml_rs::from_str(xml).unwrap();
-    data
+pub fn parse_xml(xml: &str) -> Result<OverlayData, serde_xml_rs::Error> {
+    return serde_xml_rs::from_str(xml);
 }
 
 fn convert_windows_path(path: String) -> String {
@@ -247,7 +246,10 @@ fn strip_poi_subfolder(folder: &String) -> &OsPath {
     return path;
 }
 
-pub fn process_taco_data(folder_name: String, xml_file: String) -> HashMap<u32, FinalResult> {
+pub fn process_taco_data(
+    folder_name: String,
+    xml_file: String,
+) -> Result<HashMap<u32, FinalResult>, serde_xml_rs::Error> {
     /* As per http://www.gw2taco.com/2016/01/how-to-create-your-own-marker-pack.html, two things
     1. placing pois xml file in POIs folder is valid, even if absolute paths won't be correct
     2. categorydata.xml is loaded first, and can have all the MarkerData
@@ -260,7 +262,11 @@ pub fn process_taco_data(folder_name: String, xml_file: String) -> HashMap<u32, 
     if categorydata_file.exists() {
         let default_category_content = escape_xml_str(read_to_string(categorydata_file).unwrap());
         let default_category_parsed = parse_xml(&default_category_content);
-        for marker in default_category_parsed.marker_category {
+        let default_category_parsed_unwrapped = match default_category_parsed {
+            Err(e) => return Err(e),
+            Ok(x) => x,
+        };
+        for marker in default_category_parsed_unwrapped.marker_category {
             construct_lookup_map(&mut lookup, marker, "".to_string());
         }
     }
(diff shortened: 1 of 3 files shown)


jlloh:
Ah nevermind the code isn't ready to merge. I initially tested with [this marker pack](https://fast.farming-community.eu/open-world/solo-farming) which is much smaller.

I just tried it with Tekkit's `tw_mc_coretyria.xml` which is a more complicated XML, and it fails to parse. Will attempt to fix it first.

coderedart:
@jlloh you can take some inspiration from https://github.com/coderedart/jokolay/tree/master/src/tactical/xmltypes . as long as there's no xml mistakes like wrong attribute names or duplicate fields, that should work. https://github.com/coderedart/jokolay/issues/4 should have some more info as to where i'm getting the info from.

jlloh:
Got it to work, and moved the code into the repo as a separate GDNative library.

See the results from loading Tekkit's `tw_mc_coretyria.xml` below. 

Screenshot from 2021-10-09 21-59-09

One issue though, seems that all paths are displayed, as you can see from the screenshot above.

taco_parser/src/xml_parser.rs
```diff
@@ -0,0 +1,245 @@
...
+
+#[derive(Debug, Serialize, Clone)]
+struct Icon {
+    position: [f32; 3],
+    texture: String,
+}
+
+//https://docs.rs/serde-xml-rs/0.5.1/serde_xml_rs/
+pub fn parse_xml(xml: &str) -> OverlayData {
+    let data: OverlayData = serde_xml_rs::from_str(xml).unwrap();
```
AsherGlick, line 94: I'm not going to be so strict as to say we should not have unwraps, however some of them are making errors fully opaque. I'm guessing this one is one of the culprits. If there is an error with the XML we need to show it to the user so that it can be fixed. I tried loading some of the reactif markers (https://reactif.games/taco/) which caused this module to panic, then caused burrito to crash when it tried to parse the resulting null as a json object. I think at the very minimum we need to have a `error at line #<xml_line> : <errormsg>` get bubbled up to godot so the xml files can be manually fixed if they have any errors.

AsherGlick (changes requested):

taco_parser/src/xml_parser.rs
jlloh:
```Successfully set transient property!
New Map
Saving Old Map
Loading New Map
Loading XML file from path/home/jl/hdd_drive/GW2/taco_markers/POIs/General__TheSilverwastes.xml
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { field: "missing field `MarkerCategory`" }', src/xml_parser.rs:94:57
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
ERROR: <native>: gdnative-core: method panicked (check stderr for output)
   At: src/lib.rs:37.
SCRIPT ERROR: load_taco_markers: Invalid type in function 'parse' in base '_JSON'. Cannot convert argument 1 from Nil to String.
   At: res://Spatial.gdc:297.
```
Hmm the error message does seem to get passed properly from Rust to Godot for me when I tried one of the files from Reactif (at least in so far as appearing in the terminal).
```Custom { field: "missing field `MarkerCategory`" }', src/xml_parser.rs:94:57```

Can you share the file you were trying with, so I can reproduce? And when you say error getting bubbled up to godot, do you mean you would like it to pop up as an error in godot?

File: taco_parser/README.md [hunk 2]
```diff
@@ -0,0 +1,17 @@
+# Introduction
+
+A very simple util that parses the XML and corresponding C struct trail files (using nom) from Taco (GW2 overlay), to produce a json compatible format for [Burrito](https://github.com/AsherGlick/Burrito), a GW2 overlay for Linux.
+
+Note that some sample files are included in `/data/` for unit testing.
+
+## Example used
+
+- Uses taco marker from [here](https://fast.farming-community.eu/open-world/solo-farming) as an example.
```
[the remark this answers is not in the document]
jlloh: Thanks for the feedback, have removed the files and put in a `test.xml` for the purposes of unit tests. Unfortunately I don't have a way of generating the C struct trl files, so I'm commenting out the unit test for that for now.

jlloh:
On a side note, I've realized that some marker packs, like the Reactif one, seem to follow a set of rules I was unaware of.

- MarkerCategories are separate from POIs and are in their own master file
- POIs are in the POI subfolder, so relative paths don't work

Wonder how Taco is able to parse this.

jlloh:
Ah on point 1, following on from  https://github.com/coderedart/jokolay/issues/4 led me to TacO's [original documentation](http://www.gw2taco.com/2016/01/how-to-create-your-own-marker-pack.html) which states:
 
```
categorydata.xml - this file is loaded first, it contains a default category tree. This category tree for now is pretty much only preliminary and will be expanded/rearranged later.
```

So seems I should attempt to detect if `categorydata.xml` exists and try to load that in before anything else. 

For issue (2) on relative paths and subfolders, can't really think of an elegant solution, probably will need to have some custom logic to try to check one directory above.

taco_parser/README.md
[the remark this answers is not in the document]
AsherGlick: If you want to make some sample ones without coding up a new script then you can manually build them in hex and then convert them into binary
```bash
echo "48 65 6c 6c 6f 20 57 6f 72 6c 64 0a" > input.hex
xxd -r -p input.hex output.bin
```

taco_parser/src/xml_parser.rs
AsherGlick: I believe these error are being printed directly to stdout/stderr from the rust module and not being passed into the main burrito process to be printed. Thus there is no way for burrito to be able to capture them in the future to show to the user.

jlloh:
Screenshot from 2021-10-10 22-40-24

Got it working with Reactif markers (tried with `Assistances__Explorations__CentralTyria.xml`).

AsherGlick (approved):

PR #46, opened 2021-10-17
Fixes to taco parser

jlloh:
- MarkerCategory can have an optional "texture" field as well, vs the originally assumed iconFile field that points to the trail image.

### DIFF
--- a/taco_parser/src/xml_parser_tests.rs
+++ b/taco_parser/src/xml_parser_tests.rs
@@ -94,13 +94,13 @@ mod xml_parser_tests {
         println!("{:#?}", result);
     }
 
-    #[test]
-    fn test_process_taco_data_reactif() {
-        let folder = "/home/jl/Documents/taco_packs/teh_trails".to_string();
-        let xml_file = "tt.mapcomp.xml".to_string();
-        let result = process_taco_data(folder, xml_file);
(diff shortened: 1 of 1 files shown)


taco_parser/src/xml_parser_tests.rs
```diff
@@ -90,13 +94,13 @@ mod xml_parser_tests {
...
-    // #[test]
-    // fn test_process_taco_data_reactif() {
-    //     let folder = "/home/jl/Documents/reaktif_taco/POIs/".to_string();
-    //     let xml_file = "General__Tequatl.xml".to_string();
-    //     let result = process_taco_data(folder, xml_file);
-    //     println!("{:#?}", result);
-    // }
+    #[test]
+    fn test_process_taco_data_reactif() {
+        let folder = "/home/jl/Documents/taco_packs/teh_trails".to_string();
```
AsherGlick, line 99: This looks like an absolute path to a directory on your own system. I must have missed it before or ignored it because it was in a comment. this should be a relative path so that it works on other computers.

AsherGlick (changes requested):

taco_parser/src/xml_parser_tests.rs
jlloh: Commented out, good catch.

PR #87, opened 2022-08-10
Code Generator now makes attribute files for CompoundValues and Enum

klingbolt:


xml_converter/generators/code_generator.py
```diff
@@ -368,26 +367,53 @@ def write_attribute (self, output_directory: str) -> None:
...
+                        
+                elif metadata[filepath]['type'] == "CompoundValue":
+                    
+                    for component in metadata[filepath]['components']:
+                        if component['type'] not in doc_type_to_cpp_type:
+                            raise ValueError("Unexpected type for component. Look at markdown file {attribute_name}".format(
+                                attribute_name=attribute_name
+                            ))
+                        
+                        attribute_variables.append((component['name'].lower().replace(" ","_"), doc_type_to_cpp_type[component['type']]))
```
AsherGlick, line 400: Maybe think about making a `lowercase` function to pair with the `capitalize` function below.

xml_converter/src/attribute/cull_chirality.hpp
```diff
@@ -5,13 +5,15 @@
...
 class CullChirality {
  public:
-    string chirality;
+    string cull_chirality;
+
+    virtual string classname() { return "CullChirality"; };
 };
 
-CullChirality parse_Chirality(rapidxml::xml_attribute<>* input, vector<XMLError*> *errors);
+CullChirality parse_CullChirality(rapidxml::xml_attribute<>* input, vector<XMLError*> *errors);
```
AsherGlick, line 19: I know I allowed this before because they were a technical requirement for how the , but I have changed my mind and these functions should be named in the same style as other functions. EG:

```c++
CullChirality parse_cull_chirality(rapidxml::xml_attribute<>* input, vector<XMLError*> *errors); 
```

AsherGlick (changes requested):

xml_converter/generators/code_generator.py
```diff
@@ -368,26 +367,53 @@ def write_attribute (self, output_directory: str) -> None:
         categories: Dict[str,List[str]] = {}
         attribute_names: Dict[str,str] = {}
         metadata: Dict[str, SchemaType] = {}
-         
+        attribute_variables: Tuple[List[Tuple[str, str]]] = []
+        doc_type_to_cpp_type: Dict[str,str] = {
```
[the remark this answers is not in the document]
klingbolt: I had considered that. I'll make that change

xml_converter/generators/code_generator.py
```diff
@@ -368,26 +367,53 @@ def write_attribute (self, output_directory: str) -> None:
...
+                        if component['type'] not in doc_type_to_cpp_type:
+                            raise ValueError("Unexpected type for component. Look at markdown file {attribute_name}".format(
+                                attribute_name=attribute_name
+                            ))
+                        
+                        attribute_variables.append((component['name'].lower().replace(" ","_"), doc_type_to_cpp_type[component['type']]))
                 
+                elif metadata[filepath]['type'] == "Enum":
+                    attribute_variables.append((metadata[filepath]['name'].lower().replace(" ","_"), "string"))
+                else:
```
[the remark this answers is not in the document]
klingbolt: My thought on the else was it would catch if the type field is erroneously set to multiple types.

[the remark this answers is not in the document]
AsherGlick: How would it be set to multiple types?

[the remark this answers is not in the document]
klingbolt: Ah I see. Type has to be a string so its already accounted for. Will remove the else.

xml_converter/generators/code_generator.py
klingbolt: Done

xml_converter/generators/code_generator.py
```diff
@@ -387,33 +380,29 @@ def write_attribute (self, output_directory: str) -> None:
             if metadata[filepath]['type'] in ["MultiflagValue", "CompoundValue", "Enum"]:
```
AsherGlick, line 380: This should just be

```python
                if metadata[filepath]['type'] == "MultiflagValue":
                    ...
                elif metadata[filepath]['type'] == "CompoundValue":
                    ...
                elif metadata[filepath]['type'] == "Enum":
                    ...
                else:
                    continue
                ...
```
This way we dont need to define all the string values twice.

xml_converter/src/xml_converter.cpp
```diff
@@ -218,11 +218,11 @@ void convert_taco_directory(string directory) {
 
 int main() {
     convert_taco_directory("./packs/tw_ALL_IN_ONE");
```
AsherGlick, line 220: if we want to remove these files we should just change it to look for all the files in a folder.

AsherGlick (changes requested):

xml_converter/src/xml_converter.cpp
klingbolt: I think I will do this change now so it doesn't continue to be an issue when I commit

xml_converter/generators/code_generator.py
klingbolt: Good point. I'll make that change. I didn't know about continue

AsherGlick (approved):

PR #90, opened 2022-08-22
Added a function that deletes generated files

klingbolt:
Added _gen to generated files and made adjustments nessecary to ensure functionality


xml_converter/generators/presubmit.sh
```diff
@@ -1,4 +1,4 @@
-source ./venv/bin/activate
```
AsherGlick, line 1: The virtual environment should be in `xml_converter/generators` not `xml_converter/generators/generators`. Not sure why this was changed.

AsherGlick (changes requested):

AsherGlick (approved):

Issue #71, opened 2022-07-24
[XML Markers] Export XML


AsherGlick:
The generated classes need a way to export a a file as XML. This should be done by creating a new virtual function of
```c++
string as_xml()
```
on the parseable class. Then implementing it on each of the subclasses.

For Icon and Trail the function should return the entire node along with all of the attributes it uses. Unused attributes should **NOT** be exported as attributes.

For MarkerCategory the function should export itself as a node, and if it has children then it should also export them.

PR #91, opened 2022-08-30
Added function write_xml_file() 

klingbolt:
Adds functionality to stringify attribute values from memory and save to an xml file. Addressing issue #71


xml_converter/src/xml_converter.cpp
```diff
@@ -37,10 +32,36 @@ bool has_suffix(std::string const &fullString, std::string const &ending) {
...
+void write_xml_file(string xml_filepath, map<string, Category>* marker_categories, vector<Parseable*>* parsed_pois) {
+    ofstream outfile;
+    string tab_string;
+    string file_text;
+    string new_file_path = xml_filepath + "export.xml";
+
+    file_text = "<OverlayData>\n";
+    for (const auto & category : *marker_categories) {
+        string text;
+        for (const auto& s : category.second.as_xml()) { text += s; };
```
AsherGlick, line 44: Break this out to have the inner loop on a new line

xml_converter/src/xml_converter.cpp
```diff
@@ -37,10 +32,36 @@ bool has_suffix(std::string const &fullString, std::string const &ending) {
...
+        file_text = file_text + text + "\n";
+    }
+
+    file_text = file_text + "<POIs>\n";
+    for (const auto & parsed_poi : *parsed_pois) {
+        string text; 
+        for (const auto& s : parsed_poi->as_xml()) { text += s; };
```
AsherGlick, line 51: Break this out to have the inner loop on a new line

xml_converter/src/xml_converter.cpp
```diff
@@ -37,10 +32,36 @@ bool has_suffix(std::string const &fullString, std::string const &ending) {
...
+        file_text = file_text + text + "\n";
+    }
+    file_text = file_text + "</POIs>\n";
+
+
+    outfile.open(new_file_path, ios::out);
+    outfile<< file_text;
```
AsherGlick, line 58:
```suggestion
    outfile << file_text;
```

xml_converter/src/trail_gen.hpp
```diff
@@ -29,42 +29,69 @@
...
+        string schedule;
+        bool schedule_is_true = false;
+        Color color;
+        bool color_is_true = false;
+        Image texture;
+        bool texture_is_true = false;
+        float animation_speed;
+        bool animation_speed_is_true = false;
+        bool can_fade;
+        bool can_fade_is_true = false;
```
AsherGlick, line 92: `can_fade_is_true` sounds like you are checking if the boolean `can_fade` is set to `true` which is what `can_fade` does itself. The suffix should be something that better explains what it does.

xml_converter/src/parseable.cpp
```diff
@@ -35,4 +35,10 @@ bool Parseable::init_xml_attribute(rapidxml::xml_attribute<>* attribute, vector<
     // I removed all of the offending variables. This whole section will be relooked at. 
     string item = normalize(get_attribute_name(attribute));
     return true;
-}
\ No newline at end of file
+}
+
+vector<string> Parseable::as_xml() const{
+    vector<string> result;
+    result.push_back("a");
```
AsherGlick, line 42: Lets just `panic` and freak out here instead of actually being a nicely behaved function. Look into something like `assert(false)` or throwing an error or `exit(123245)` or something.

xml_converter/src/icon_gen.cpp
```diff
@@ -8,213 +10,456 @@ string Icon::classname() {
...
+        this->z_rotation = parse_float(attribute, errors);
+        this->z_rotation_is_true = true;
     }
-    else if (attributename == "inforange") {
-        this->trigger_range = parse_float(attribute, errors);
+    else if (attributename == "rotatez") {
+        this->euler_rotation = parse_euler_rotation(attribute, errors);
+        this->euler_rotation_is_true = true;
+    }
+    else if (attributename == "rotate") {
```
AsherGlick, line 293: You have `rotate` multiple times here

xml_converter/src/attribute/bool.cpp
```diff
@@ -21,3 +21,16 @@ bool parse_bool(rapidxml::xml_attribute<>* input, vector<XMLError*> *errors) {
         return false;
     }
 }
+
+string stringify_bool(bool attribute_value){
+    if (attribute_value == false) {
```
AsherGlick, line 26: We can just do "if" and "else" because  bool can only be true or false

xml_converter/generators/code_generator.py
```diff
@@ -333,7 +337,8 @@ def write_cpp_classes(self, output_directory: str) -> None:
             with open(os.path.join(output_directory, lowercase(cpp_class) + "_gen.hpp"), 'w') as f:
                 f.write(header_template.render(
                     cpp_class=cpp_class,
-                    attribute_variables=sorted(attribute_variables, key=get_attribute_variable_key),
+                    # attribute_variables=sorted(attribute_variables, key=get_attribute_variable_key),
```
AsherGlick, line 340: We should still sort stuff.

xml_converter/generators/code_generator.py
```diff
@@ -343,7 +348,8 @@ def write_cpp_classes(self, output_directory: str) -> None:
                     cpp_class=cpp_class,
                     cpp_class_header=lowercase(cpp_class),
                     xml_class_name=cpp_classes[cpp_class],
-                    attribute_variables=sorted(attribute_variables, key=get_attribute_variable_key),
+                    # attribute_variables=sorted(attribute_variables, key=get_attribute_variable_key),
```
AsherGlick, line 351: We should still sort stuff.

AsherGlick (changes requested):

xml_converter/src/xml_converter.cpp
klingbolt: Done

xml_converter/src/xml_converter.cpp
klingbolt: Done

xml_converter/src/xml_converter.cpp
klingbolt: Done

xml_converter/src/trail_gen.hpp
klingbolt: Changed to ```is_set```

xml_converter/src/parseable.cpp
klingbolt: Exit failure added. Function still needs to return something to compile.

xml_converter/src/icon_gen.cpp
klingbolt: Found the logic error. Components now have a different variable name in the loop so the value cannot escape.

xml_converter/src/attribute/bool.cpp
klingbolt: Done

xml_converter/generators/code_generator.py
klingbolt: Sorting has been restored.

xml_converter/generators/code_generator.py
klingbolt: Sorting has been restored.

xml_converter/src/attribute/bool.cpp
AsherGlick: Lets actually change this to `if (attribute_value) { return "true" } else { return "false"}` (with proper spacing.

xml_converter/src/parseable.cpp
```diff
@@ -38,7 +38,9 @@ bool Parseable::init_xml_attribute(rapidxml::xml_attribute<>* attribute, vector<
 }
 
 vector<string> Parseable::as_xml() const{
+    cout << "error: Parseable::as_xml() should not be called";
+    exit(EXIT_FAILURE);
     vector<string> result;
-    result.push_back("a");
+    result.push_back("");
```
AsherGlick, line 44: No need to have this `push_back`. Even though none of this will ever be called.

xml_converter/src/parseable.cpp
```diff
@@ -38,7 +38,9 @@ bool Parseable::init_xml_attribute(rapidxml::xml_attribute<>* attribute, vector<
 }
 
 vector<string> Parseable::as_xml() const{
+    cout << "error: Parseable::as_xml() should not be called";
+    exit(EXIT_FAILURE);
```
AsherGlick, line 42: Lets use exceptions instead of exit here. This is so we can catch them in a test case.

xml_converter/src/attribute/bool.cpp
klingbolt: Done

xml_converter/src/parseable.cpp
klingbolt: Removed

xml_converter/src/xml_converter.cpp
```diff
@@ -37,10 +32,40 @@ bool has_suffix(std::string const &fullString, std::string const &ending) {
...
+        file_text = file_text + text + "\n";
+    }
+    file_text = file_text + "</POIs>\n";
+
+
+    outfile.open(new_file_path, ios::out);
+    outfile << file_text;
+    outfile.close();
+}
+
```
AsherGlick, line 65:
```suggestion
void write_xml_file(string xml_filepath, map<string, Category>* marker_categories, vector<Parseable*>* parsed_pois) {
    ofstream outfile;
    string tab_string;
    string new_file_path = xml_filepath + "export.xml";

    outfile.open(new_file_path, ios::out);

    outfile << "<OverlayData>\n";
    for (const auto & category : *marker_categories) {
        string text;
        for (const auto& s : category.second.as_xml()) { 
            text += s;
        }
        outfile << text + "\n";
    }

    outfile << "<POIs>\n";
    for (const auto & parsed_poi : *parsed_pois) {
        string text; 
        for (const auto& s : parsed_poi->as_xml()) { 
            text += s; 
        }
        outfile << text + "\n";
    }
    outfile << "</POIs>\n";

    outfile.close();
}
```

AsherGlick (approved):

Issue #72, opened 2022-07-24
[XML Markers] Internal Protobuf Schema


AsherGlick:
A protobuf schema needs to be written to handle the internal filetypes.

It is possible it can be auto generated from the docs, though there might be some file types that are difficult to store. Those custom file types may need to be revisited to see if they can be dynamic instead. Additionally if generated automatically then a mapping between internal types and protobuf types would need to exist somewhere.

Another option, and the suggested one, is to write the schema manually and rely on C++ compiler errors to validate the schema complies with all of the fields required. This will allow custom data types to implement their own serializer functions without needing tooling for bespoke automatic generation.

Issue #58, opened 2022-04-03
Add EOD Spec Values to XML Parser 


AsherGlick:
`xml_converter/src/attribute/specialization_filter.hpp` needs the EOD elite specs added to its list to be able to parse them properly.

PR #95, opened 2022-10-17
Wrote Internal Protobuf Schema for attributes

klingbolt:
As requested in issue #72 

Included the addition of EOD spec values as requested in issue #58


xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,274 @@
...
+    bool __tentative__render_on_minimap = 15;
+    bool scale_on_map_with_zoom = 16;
+    string bhdraft__schedule = 17;
+    float bhdraft__schedule_duration = 18;
+    string tip_description = 19;
+    string tip_name = 20;
+    
+    message texture {
+        string path = 1;
+        google.protobuf.Any original_token = 2;
```
klingbolt, line 39: I'm uncertain on this typing. The program currently uses a pointer in C++.

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,274 @@
...
+    bool __tentative__render_ingame = 13;
+    bool __tentative__render_on_map = 14;
+    bool __tentative__render_on_minimap = 15;
+    bool scale_on_map_with_zoom = 16;
+    string bhdraft__schedule = 17;
+    float bhdraft__schedule_duration = 18;
+    string tip_description = 19;
+    string tip_name = 20;
+    
+    message texture {
```
AsherGlick, line 37: Don't use nested messages here. They are not nested in the python code which means that these will just end up being global types that get duplicate nesting.

Just make them global messages.

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,274 @@
...
+
+message Icon {
+    fixed32 achievement_bit = 1;
+    int32 achievement_id = 2;
+    float alpha = 3;
+    bool can_fade = 4;
+    float distance_fade_end = 5;
+    float distance_fade_start = 6;
+    float height_offset= 7;
+    float __tentative__scale = 8;
```
AsherGlick, line 23: A `__tentative__`, `bhdraft__` or otherwise not finalized value should not be given a single byte index. This should probably be expanded to at least a double byte but probably a triple byte index instead.

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,274 @@
...
+    bool default_visibility = 1;
+    string display_name = 2;
+    bool is_separator = 3;
+    string name = 4;
+    string tooltip_name = 5;
+    map <string, Category> children = 6;
+}
+
+message Icon {
+    fixed32 achievement_bit = 1;
```
AsherGlick, line 16: Lower indexes should be reserved for more commonly used items. I would expect the first few numbers to be things like `map_id`, and `position`.

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,274 @@
+syntax = "proto3";
+import "google/protobuf/any.proto";
```
AsherGlick, line 2: Referencing the later comment. `Any` should not be necessary so we probably don't need this.

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,274 @@
...
+
+package Proto_Node;
```
AsherGlick, line 4: We are not generating java code so we can omit the `package` element here. Better to leave it not defined in case we ever need to define it in the future.

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,274 @@
...
+message Category {
+    bool default_visibility = 1;
+    string display_name = 2;
+    bool is_separator = 3;
+    string name = 4;
+    string tooltip_name = 5;
+    map <string, Category> children = 6;
+}
+
+message Icon {
```
AsherGlick, line 15: I think you are using nested message type definitions and thinking they are variable definitions.

xml_converter/generators/cpp_templates/class_template.hpp
```diff
@@ -36,11 +38,13 @@ class {{cpp_class}}: public Parseable {
         Icon default_icon;
         Trail default_trail;
 
+        virtual vector<string> as_xml() const;
         void init_from_xml(rapidxml::xml_node<>* node, vector<XMLError*> *errors);
+        {%- else %}
+        virtual vector<string> as_xml() const;
```
AsherGlick, line 44: Not sure why there is explicitly duplicated logic now
`if cpp_class == "Category"` you include `virtual vector<string> as_xml() const`
`else` you still include `virtual vector<string> as_xml() const;`
meaning that in all cases it is included.

This change should be undone.

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,274 @@
...
+        bool engineer = 3;
+        bool ranger = 4;
+        bool thief = 5;
+        bool elementalist = 6;
+        bool mesmer = 7;
+        bool necromancer = 8;
+        bool revenantnt = 9;
+    }
+    message specialization_filter {
+    // Heart of Thorns Spec
```
AsherGlick, line 178: While clever to use the spec ids it does cause some level of confusion when presented in this proto file where most fields are ordered by their index. It's fine to continue using the spec ids here but they should be ordered by their id and any possibly future or past unused id that should not be used should be marked as `reserved`. Otherwise it is possible for arenanet to allocate and id and for a future revision of this code to have allocated an id, and then for those ids to collide causing this clever alignment to break down.

AsherGlick (changes requested):

xml_converter/generators/cpp_templates/class_template.hpp
klingbolt: Eventually as_xml() will be unique for category because the number of tabs needs to be passed in and out.

Since I reverted my experiments on that function, I will revert it here.

xml_converter/generators/proto_templates/node.proto
klingbolt: Removed

xml_converter/generators/proto_templates/node.proto
klingbolt: Removed.

xml_converter/generators/proto_templates/node.proto
klingbolt: Reworked to avoid nested messages

xml_converter/generators/proto_templates/node.proto
klingbolt: Reworked indexes

xml_converter/generators/proto_templates/node.proto
klingbolt: Rework to be at index 2048 and up.

xml_converter/generators/proto_templates/node.proto
klingbolt: Reworked to avoid Nested messages

xml_converter/generators/proto_templates/node.proto
klingbolt: As discussed, this variable is unused in the code and has been removed

xml_converter/generators/proto_templates/node.proto
klingbolt: Understood. Reworked indexes

xml_converter/generators/proto_templates/node.proto
```diff
@@ -13,262 +10,281 @@ message Category {
...
+    bool lunar_new_year = 4;
+    bool super_adventure_festival = 5;
+    bool wintersday = 6;
+    bool none = 7;
+}
 
-    message guid{
-        int32 guid = 1;
-    }
+message Guid{
```
AsherGlick, line 140: You have `Guid` here and `GUID` above.

xml_converter/generators/proto_templates/node.proto
```diff
@@ -13,262 +10,281 @@ message Category {
...
+    bool has_countdown = 7;
+    string action_info_message = 8;
+    bool invert_display = 9;
+    float reset_length = 10;
+    float range = 11;
+    Category action_hide_category = 12;
+    Category action_show_category = 13;
+    Category action_toggle_category = 14;
 
+message GUID{
```
AsherGlick, line 105: You have `GUID` here and `Guid` below.

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,290 @@
...
+    bool thief_shadow_arts = 66;
+    bool thief_trickery = 67;
+    bool warrior_arms = 68;
+    bool warrior_defense = 69;
+    bool warrior_discipline = 70;
+    bool warrior_strength = 71;
+    bool warrior_tactics = 72;
+}
+
+message Species_filter {
```
AsherGlick, line 275: This should probably be `SpeciesFilter`

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,290 @@
...
+    bool human = 3;
+    bool norn = 4;
+    bool sylvari = 5;
+}
+    
+message Texture {
+    string path = 1;
+}
+
+message Trail_data {
```
AsherGlick, line 287: this should probably be `TrailData`

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,290 @@
...
+    bool ranger = 4;
+    bool thief = 5;
+    bool elementalist = 6;
+    bool mesmer = 7;
+    bool necromancer = 8;
+    bool revenantnt = 9;
+}
+
+message Specialization_filter {
+// Heart of Thorns Spec
```
AsherGlick, line 197: Nit: Lets indent these comments.

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,290 @@
...
+    bool jackal = 4;
+    bool griffon = 5;
+    bool roller_beetle = 6;
+    bool warclaw = 7;
+    bool skyscalee = 8;
+    bool skiff = 9;
+    bool seige_turtle = 10;
+}
+
+message Profession_filter {
```
AsherGlick, line 184: This should probably be `ProfessionFilter`

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,290 @@
...
+    bool engineer = 3;
+    bool ranger = 4;
+    bool thief = 5;
+    bool elementalist = 6;
+    bool mesmer = 7;
+    bool necromancer = 8;
+    bool revenantnt = 9;
+}
+
+message Specialization_filter {
```
AsherGlick, line 196: This should probably be `SpecializationFilter`

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,290 @@
...
+    bool red_borderlands_map = 18;
+    bool fortunes_vale_map = 19;
+    bool jump_puzzle_map = 20;
+    bool obsidian_sanctum_map = 21;
+    bool edge_of_the_mists_map = 22;
+    bool public_mini_map = 23;
+    bool wvw_lounge_map = 24;
+}
+
+message Mount_filter {
```
AsherGlick, line 171: This should probably be `MountFilter`

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,290 @@
...
+    bool super_adventure_festival = 5;
+    bool wintersday = 6;
+    bool none = 7;
+}
+
+message Guid{
+    int32 guid = 1;
+}
+
+message Map_type_filter {
```
AsherGlick, line 144: This should probably be `MapTypeFilter`

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,290 @@
...
+        daily_reset = 2;
+        never = 3;
+        timer = 4;
+        map_reset = 5;
+        instance_change = 6;
+        daily_reset_per_character = 7;
+        weekly_reset = 8;
+}
+
+message Festival_filter {
```
AsherGlick, line 130: This should probably be `FestivalFilter`

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,290 @@
...
+message Color {
+    string hex = 1;
+}
+
+enum Cull_chirality {
+        none = 0;
+        clockwise = 1;
+        counter_clockwise = 2;
+}
+enum reset_behavior {
```
AsherGlick, line 118: This should probably be `ResetBehavior`

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,290 @@
...
+
+message GUID{
+    int32 guid = 1;
+}
+
+message Color {
+    string hex = 1;
+}
+
+enum Cull_chirality {
```
AsherGlick, line 113: This should probably be `CullChirality`

xml_converter/generators/proto_templates/node.proto
```diff
@@ -0,0 +1,290 @@
...
+    string path = 1;
+}
+    
+message Position {
+    float x = 1;
+    float y = 2;
+    float z = 3;
+}
+
+message Euler_rotation {
```
AsherGlick, line 83: This should probably be `EulerRotation`

xml_converter/generators/cpp_templates/class_template.hpp
AsherGlick: While each `as_xml()` function logic will be unique, the function signatures should all be identical.

AsherGlick (changes requested):

AsherGlick (approved):

Issue #73, opened 2022-07-24
[XML Markers] Protoc and C++ Dependencies


AsherGlick:
Protoc can generate c++ protobuf code, however that code is dependent on external libraries. Those libraries need to be included in the compilation commands, additionally those libraries will need to be included in the CI system to compile these binaries so we can have proper automated builds of this new library.

PR #96, opened 2022-11-03
Protoc Dependencies added to CMake file

klingbolt:
Addressing issue #73


xml_converter/CMakeLists.txt
```diff
@@ -6,13 +6,22 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 # Name Output File
 set(TARGET_NAME xml_converter)
 
+# Add Protobuf
+find_package(Protobuf REQUIRED)
+include_directories(${Protobuf_INCLUDE_DIRS})
+# include_directories()
+EXECUTE_PROCESS(COMMAND ${PROTOBUF_PROTOC_EXECUTABLE} --proto_path=generators/proto_templates --cpp_out=src/ generators/proto_templates/waypoint.proto)
```
AsherGlick, line 13: `EXECUTE_PROCESS` is run when `cmake` itself is called. The protobuf generation should probably be run when `make` is run instead of when `cmake` is run.

Additionally these paths do not function correctly when making an "out of source" build, which is generally recommended to do. An example of an out of source build is:

```bash
mkdir build
cd build
cmake ..
make
./my_executable
```

This means that when the build process is complete you can just run

```bash
rm -rf build
```
and clean up all the changes made by the build process. In our system things are a tad bit weird because we commit the code we generated from the generator we wrote into the source directory, but I think that is fine because the generator itself is a part of this codebase. Maybe in the future once this is strongly tested the generated files themselves will live outside the codebase and they will also be generated by the makefile just like the proto code.

AsherGlick (changes requested):
Don't include the generated protobuf code in the repo. Depending on the protoc version installed on the compiling machine the generated code will be different. Our own generated code is a small exception right now because we also have the generator in the repo. But as mentioned in another comment we might want to remove our own generated code in the future as well, but not until all this code is well battle tested in production and tests.

.github/workflows/main.yml
```diff
@@ -89,6 +89,17 @@ jobs:
             cd taco_parser
             cargo build --release
 
+      - name: Install protoc
+        run: sudo apt-get install protobuf-compiler 
+
+      - name: Build xml_converter
+        run: |
+            cd xml_converter
+            protoc --proto_path=generators/proto_templates --cpp_out=src/ generators/proto_templates/waypoint.proto
```
klingbolt, line 98: This works as intended but I'm curious if there is a best practice. Since the Makefile for xml_convertor is not included in the code, cmake needs to be run. Therefore I thought it'd be best for the proto buffers to be generated here. Similarly I could add a line to run code_generator.py

AsherGlick: This is the equivalent of "please manually run `protoc`" which is not a good build experience. Protoc should be run inside the generated makefile as originally requested.

In general there is no issue with putting commands in the CI script, however in this  particular case it should be a part of the makefile. Right now if you try to type `cmake . && make` it would crash on compilation because the generated proto files would not exist.

AsherGlick (changes requested):

xml_converter/proto/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+INCLUDE(FindProtobuf) 
+FIND_PACKAGE(Protobuf REQUIRED)
```
AsherGlick, line 2: Do you need `FIND_PACKAGE(Protobuf REQUIRED)` both here and in the main CMake file? Or can one of them be eliminated?

AsherGlick (changes requested):

xml_converter/generators/cpp_templates/class_template.hpp
```diff
@@ -40,8 +40,8 @@ class {{cpp_class}}: public Parseable {
 
         void init_from_xml(rapidxml::xml_node<>* node, vector<XMLError*> *errors);
         {%- else %}
```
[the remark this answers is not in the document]
klingbolt: No it is not needed at this time. Removed

xml_converter/proto/CMakeLists.txt
klingbolt: I did tests and the FIND_PACKAGE line can be removed. INCLUDE is required.

xml_converter/proto/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+INCLUDE_DIRECTORIES(${PROTOBUF_INCLUDE_DIR})
+PROTOBUF_GENERATE_CPP(PROTO_SRC PROTO_HEADER waypoint.proto)
+ADD_LIBRARY(proto ${PROTO_HEADER} ${PROTO_SRC})
```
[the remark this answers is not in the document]
klingbolt: That makes sense. I will keep the generated files in the /proto file and change the references

xml_converter/proto/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+
+add_custom_command(TARGET proto PRE_BUILD
```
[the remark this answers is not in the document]
klingbolt: I will keep the generated files in the /proto file and change the references

xml_converter/src/xml_converter.cpp
```diff
@@ -18,6 +18,7 @@
 #include "attribute/float.hpp"
 #include "string_helper.hpp"
 #include "rapid_helpers.hpp"
+#include "waypoint.pb.h"
```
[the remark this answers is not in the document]
klingbolt: Adding a test function makes sense to me.

.github/workflows/main.yml
```diff
@@ -89,6 +89,16 @@ jobs:
             cd taco_parser
             cargo build --release
 
+      - name: Install protoc
+        run: sudo apt-get install protobuf-compiler 
+
+      - name: Build xml_converter
+        run: |
+            cd xml_converter
```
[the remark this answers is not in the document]
klingbolt: done

xml_converter/generators/code_generator.py
```diff
@@ -220,6 +221,9 @@ def validate_front_matter_schema(front_matter: Any) -> str:
         return "Error Message: {} (Path: {}".format(e.message, e.json_path)
     return ""
 
+# def validate_protobuf_schema()
```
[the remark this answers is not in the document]
klingbolt: done

xml_converter/proto/CMakeLists.txt
klingbolt, line 2:
```suggestion
```

xml_converter/proto/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+INCLUDE(FindProtobuf) 
+INCLUDE_DIRECTORIES(${PROTOBUF_INCLUDE_DIR})
+PROTOBUF_GENERATE_CPP(PROTO_SRC PROTO_HEADER waypoint.proto)
+ADD_LIBRARY(proto ${PROTO_HEADER} ${PROTO_SRC})
+
+add_custom_command(TARGET proto PRE_BUILD
+    COMMAND ${CMAKE_COMMAND} -E copy ${PROTO_HEADER} ${PROTO_SRC} ../../src)
+
+add_custom_command(TARGET proto POST_BUILD
+    COMMAND ${CMAKE_COMMAND} -E rm ${PROTO_HEADER} ${PROTO_SRC})
```
klingbolt, line 10:
```suggestion
```

AsherGlick (approved):

PR #97, opened 2022-11-06
Performance and linting

AsherGlick:
Various changes to increase performance, and linter checks to decrease errors. Linter checks are not enforced at this time.

### DIFF
--- a/xml_converter/src/string_helper.cpp
+++ b/xml_converter/src/string_helper.cpp
@@ -118,11 +118,6 @@ static const std::string base64_chars =
              "abcdefghijklmnopqrstuvwxyz"
              "0123456789+/";
 
-
-static inline bool is_base64(uint8_t c) {
-  return (isalnum(c) || (c == '+') || (c == '/'));
-}
-
 std::string base64_encode(uint8_t const* buf, unsigned int bufLen) {
   std::string ret;
   int i = 0;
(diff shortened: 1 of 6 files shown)


xml_converter/src/string_helper.cpp
```diff
@@ -93,11 +118,6 @@ static const std::string base64_chars =
              "abcdefghijklmnopqrstuvwxyz"
```
klingbolt, line 118: Since the functions here have been changed, should the attribution be removed?

xml_converter/src/string_helper.cpp
```diff
@@ -142,49 +162,83 @@ std::string base64_encode(uint8_t const* buf, unsigned int bufLen) {
   return ret;
 }
 
+#include "function_timers.hpp"
+
+// TODO: Write Tests For This
+static unsigned char base64_lookup[256] = {
```
klingbolt, line 168: This table mirrors the one used in the normalize_lookup. Does it follow the same mapping as in that table?

xml_converter/src/string_helper.cpp
AsherGlick: I dont believe I modified the `base64_encode` function so leaving the attribution there is fine it just only references the encode function now instead of both the encode and decode. Mildly confusing maybe but easily discoverable upon inquiry by clicking the link.

xml_converter/src/string_helper.cpp
AsherGlick: This absolutely does **NOT** mirror the `normalize_lookup` table. This table maps an ASCII character to an integer value or an error value (255), `normalize_lookup` maps an ASCII character to another ASCII character or null/ignore value. This table supports all base64 characters which include two symbols `+` and `/`. Those characters are ignored in the normalize table.

klingbolt: Can you please add a comment in the code similar to line 59 that describes the mapping for this table?

AsherGlick: Done

PR #105, opened 2022-11-14
Cpp linting in CI

AsherGlick:
Adding a set of linting tools for the code in the xml_converter directory. Enabling it to be run as part of the CI process.


AsherGlick:
The linter and CI systems seem to work well now. However before merging this in, the generators should be updated to output code that the linters accept.

klingbolt (approved):

Issue #75, opened 2022-07-24
[XML Markers] Output Protobins Using Internal Category and Map Hierarchy


AsherGlick:
Once the Protobuf Classes have been populated we need a way to split them up and write them to the internal file structure. Each category will be split up by map id so that the user only needs to load markers for the particular map they are on. Additionally we want to allow for an easy construction of the hierarchy of what categories are in each map. Therefore the internal map hirearchy will look like this
```
<map_id>/
    base_category/
        sub_category_1/
            sub_category_1_1.bin
            sub_category_1_2.bin
        sub_category_2/
            sub_category_2_1.bin
        sub_category_2.bin
```
This will allow burrito to simply list all the files in a directory in order to later figure out what are all the categories in that directory. if we add the ability to show *all* categories when a user wants to see them we can then just create a single file that has a map of all the categories, instead of having Burrito parse every <map_id> folder

coderedart:
o/ long time no see.

I almost went with a similar approach at first, but this layout will create SO MANY FILES. 

tekkit for example has over 8k categories https://mp-repo.blishhud.com/tw_ALL_IN_ONE.taco.html
and some of those categories will be duplicated across maps, so its even more files within a single marker pack. and an equivalent number of categories. 

at the same time, having your category menu tree spread out across map folders means that category menu can be different per different maps. also, there's no defined "order" of category menu items, so for example in a queensdale map
```
15/ 
	guild_missions/
	map_completion/
	achievements/
```
how does someone define the categories to appear in menu in that order with just (sub-)directories?	

> there's roughly less than a thousand maps https://api.guildwars2.com/v2/maps and most of them are never going to have markers. tekkit only has markers on 187 maps (refer above link), so 200 ish files for the largest pack doesn't sound bad.

AsherGlick:
The internal binary format is not intended for use by humans. Users will only need to care about the "full" category list, and not individual map categories. However only categories that are present in the map you are in will show up by default in the category selection, as to cut down on menu size. With this design Burrito can just query the directory and subdirectories to find what categories are on the map because if a map does not have any markers for a given category no folder or file should exist. 

Category ordering though is a very good point. Directories do not have a set ordering so that data could not be properly preserved under this design. @coderedart Do you have a suggustion as to a different method of storing the files internally?

coderedart:
Even if its internally, creating thousands of files / directories for a single marker pack doesn't feel warranted (especially when there's no compelling reason to do so). 
[...]

PR #114, opened 2022-12-20
Output Protobins Using Map Identification 

klingbolt:
Solving issue #75 . The code creates a `./protobins` folder and serializes a separate protobin file for each map ID containing only the categories and icons present in that map.

### DIFF
--- a/xml_converter/src/xml_converter.cpp
+++ b/xml_converter/src/xml_converter.cpp
@@ -65,28 +65,7 @@ void write_xml_file(string xml_filepath, map<string, Category>* marker_categorie
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-// create_proto_categories
-//
-// Fills a set of the names of categories for validation purposes
-// eg.
-// {
-//     "mypath",
-//     "easypath",
-//     "trail",
-//     "hardpath",
-// }
-////////////////////////////////////////////////////////////////////////////////
-void create_proto_categories(waypoint::Category proto_category, set<string>* proto_categories) {
-    proto_categories->insert(lowercase(proto_category.name()));
-    for (int i = 0; i < proto_category.children_size(); i++) {
-        create_proto_categories(proto_category.children(i), proto_categories);
-    }
-}
-
-////////////////////////////////////////////////////////////////////////////////
-// remove_proto_child
-//
-// Compares the relative name of the category against a set of the names of POIS
+// Adds the name of a category and all of its children to a setgit 
 // eg.
 // {
 //     "mypath",
@@ -95,6 +74,37 @@ void create_proto_categories(waypoint::Category proto_category, set<string>* pro
 //     "mypath.hardpath",
 //     "mypath.hardpath.trail",
 // }
+////////////////////////////////////////////////////////////////////////////////
+void add_children_names(waypoint::Category proto_category, set<string>* proto_categories, string parent_name) {
+    string name = parent_name + "." + lowercase(proto_category.name());
+    proto_categories->insert(name);
+    for (int i = 0; i < proto_category.children_size(); i++) {
+        add_children_names(proto_category.children(i), proto_categories, name);
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// Adds the name of a category and all of it's parents to a set
+////////////////////////////////////////////////////////////////////////////////
+void add_parent_names(string category_name, set<string> proto_categories, set<string>* category_includes) {
+    auto category = proto_categories.find(category_name);
+    if (category == proto_categories.end()) {
+        cout << "Unknown MarkerCategory " << category_name << endl;
+    }
+    else {
+        string name;
+        vector<string> split_categories = split(category_name, ".");
+        for (unsigned int j = 0; j < split_categories.size(); j++) {
+            name += split_categories[j];
+            category_includes->insert(name);
+            name += ".";
+        }
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// Iterates through all children of a Category and removes those that do not
+// have POIs with corresponding names
 ////////////////////////////////////////////////////////////////////////////////
 void remove_proto_child(waypoint::Category* proto_category, set<string> category_includes, string parent_name) {
     int keep = 0;
@@ -121,15 +131,15 @@ void write_protobuf_file(string proto_filepath, map<string, Category>* marker_ca
         cout << "Error making ./protobins" << endl;
         throw std::error_code();
     }
-    // Converts map<string, Category> to a Waypoint::Category
+    // Collects a set of category names and their children
     waypoint::Waypoint all_categories;
     std::set<string> proto_categories;
     for (const auto& category : *marker_categories) {
         waypoint::Category proto_category = category.second.as_protobuf();
         all_categories.add_category()->CopyFrom(proto_category);
-        create_proto_categories(proto_category, &proto_categories);
+        add_children_names(proto_category, &proto_categories, proto_category.name());
     }
-    // Collects a set of map ids from vector<Parseable*>
+    // Collects a set of map ids from Icon and Trail data
     std::set<int> map_ids;
     for (const auto& parsed_poi : *parsed_pois) {
         if (parsed_poi->classname() == "POI") {
@@ -147,58 +157,29 @@ void write_protobuf_file(string proto_filepath, map<string, Category>* marker_ca
         ofstream outfile;
         string output_filepath = proto_filepath + "/" + to_string(map_id) + ".data";
         outfile.open(output_filepath, ios::out | ios::binary);
-        output_message.CopyFrom(all_categories);
+        output_message.MergeFrom(all_categories);
 
         for (const auto& parsed_poi : *parsed_pois) {
             if (parsed_poi->classname() == "POI") {
                 Icon* icon = dynamic_cast<Icon*>(parsed_poi);
                 if (icon->map_id == map_id) {
-                    waypoint::Icon proto_icon = icon->as_protobuf();
-                    vector<string> split_categories = split(proto_icon.category().name(), ".");
-                    auto category = proto_categories.find(lowercase(split_categories[split_categories.size() - 1]));
-                    if (category == proto_categories.end()) {
-                        cout << "Unknown MarkerCategory " << proto_icon.category().name() << endl;
-                    }
-                    else {
-                        // This loop adds the name of the category and all of it's parents
-                        string name;
-                        for (unsigned int j = 0; j < split_categories.size(); j++) {
-                            name += split_categories[j];
-                            category_includes.insert(name);
-                            name += ".";
-                        }
-                    }
-                    output_message.add_icon()->CopyFrom(proto_icon);
+                    add_parent_names(icon->category.category, proto_categories, &category_includes);
+                    output_message.add_icon()->MergeFrom(icon->as_protobuf());
                 }
             }
             else if (parsed_poi->classname() == "Trail") {
                 Trail* trail = dynamic_cast<Trail*>(parsed_poi);
                 if (trail->map_id == map_id) {
-                    waypoint::Trail proto_trail = trail->as_protobuf();
-                    vector<string> split_categories = split(proto_trail.category().name(), ".");
-                    auto category = proto_categories.find(lowercase(split_categories[split_categories.size() - 1]));
-                    if (category == proto_categories.end()) {
-                        cout << "Unknown MarkerCategory " << proto_trail.category().name() << endl;
-                    }
-                    else {
-                        // This loop adds the name of the category and all of it's parents
-                        string name;
-                        for (unsigned int j = 0; j < split_categories.size(); j++) {
-                            name += split_categories[j];
-                            category_includes.insert(name);
-                            name += ".";
-                        }
-                    }
-                    output_message.add_trail()->CopyFrom(proto_trail);
+                    add_parent_names(trail->category.category, proto_categories, &category_includes);
+                    output_message.add_trail()->MergeFrom(trail->as_protobuf());
                 }
             }
         }
-        // In the XML, Marker_Categories is a tree hierarchy while
-        // POIS have a flat hierarchy.
-        // This is preserved in the protobuf for ease of translation.
-        // We are doing a removal instead of an insertion because each parent
-        // category contains the data for all of its children.
-        // It would be impractical to include all of the data from each category
+        // In the XML, MarkerCategories have a tree hierarchy while POIS have a
+        // flat hierarchy. This is preserved in the protobuf for ease of
+        // translation. We are doing a removal instead of an insertion because
+        // each parent category contains the data for all of its children. It
+        // would be impractical to include all of the data from each category
         // except for the children and then iterating over all the children.
         // This pruning method is slower but ensures that information is kept.
         for (int i = 0; i < output_message.category_size(); i++) {
--- a/xml_converter/src/xml_converter.cpp
+++ b/xml_converter/src/xml_converter.cpp
@@ -76,7 +76,7 @@ void write_xml_file(string xml_filepath, map<string, Category>* marker_categorie
 //     "mypath.hardpath.trail",
 // }
 ////////////////////////////////////////////////////////////////////////////////
-void populate_categories_to_retain (string category_name, set<string>* categories_to_retain) {
+void populate_categories_to_retain(string category_name, set<string>* categories_to_retain) {
     string name;
     vector<string> split_categories = split(category_name, ".");
     for (unsigned int j = 0; j < split_categories.size(); j++) {
--- a/xml_converter/src/xml_converter.cpp
+++ b/xml_converter/src/xml_converter.cpp
@@ -1,4 +1,5 @@
 #include <dirent.h>
+#include <errno.h>
 #include <sys/stat.h>
 
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


xml_converter/src/xml_converter.cpp
```diff
@@ -61,30 +63,115 @@ void write_xml_file(string xml_filepath, map<string, Category>* marker_categorie
...
+    for (int map_id : map_ids) {
+        ofstream outfile;
+        string output_filepath = proto_filepath + "/" + to_string(map_id) + ".data";
+        outfile.open(output_filepath, ios::out | ios::binary);
+        output_message.CopyFrom(all_categories);
+
+        for (int i = 0; i < proto_pois.icon_size(); i++) {
+            vector<string> split_categories = split(proto_pois.icon(i).category().name(), ".");
+            if (proto_pois.icon(i).map_id() == map_id) {
+                auto category = proto_categories.find(split_categories[split_categories.size() - 1]);
```
AsherGlick, line 137: Will this return a false positive if there is a category such as
`mypack.hardpath.trail`

and you are looking for
`mypack.easypath.trail`

xml_converter/src/xml_converter.cpp
```diff
@@ -61,30 +63,115 @@ void write_xml_file(string xml_filepath, map<string, Category>* marker_categorie
...
+                if (category == proto_categories.end()) {
+                    cout << "Unknown MarkerCategory " << proto_pois.icon(i).category().name() << endl;
+                }
+                else {
+                    for (string split_category : split_categories) {
+                        category_includes.insert(split_category);
+                    }
+                }
+                output_message.add_icon()->CopyFrom(proto_pois.icon(i));
```
AsherGlick, line 146: In a future step we will need to think about how to handle this case visually. How does one hide/show something without a category. We might want to create missing categories so that they can be shown and hidden. This should probably not happen here though.

It brings up the interesting point of: if every POI or Trail has a category, should the cateogry proto just own them instead of having a separate array. This would change the proto structure pretty significantly.

xml_converter/src/xml_converter.cpp
```diff
@@ -61,30 +63,115 @@ void write_xml_file(string xml_filepath, map<string, Category>* marker_categorie
...
+
 void write_protobuf_file(string proto_filepath, map<string, Category>* marker_categories, vector<Parseable*>* parsed_pois) {
-    ofstream outfile;
-    waypoint::Waypoint proto_message;
+    waypoint::Waypoint proto_pois;
+    waypoint::Waypoint all_categories;
+    waypoint::Waypoint output_message;
+    std::set<int> map_ids;
+    std::set<string> proto_categories;
+    std::set<string> category_includes;
```
AsherGlick, line 100: I am not a fan of defining all the variables at the start of a function, it makes it difficult to see where they are used. Very old versions of C require this. These should be moved to more logical locations such as before whatever blocks need them, or if they don't need to be scoped to the function: within whatever block scope makes sense.

Of course there are times where the most logical place for some variables to be defined is indeed the top of the  function, but this is not one of them.

xml_converter/src/xml_converter.cpp
```diff
@@ -61,30 +63,115 @@ void write_xml_file(string xml_filepath, map<string, Category>* marker_categorie
...
+                else {
+                    for (string split_category : split_categories) {
+                        category_includes.insert(split_category);
+                    }
+                }
+                output_message.add_trail()->CopyFrom(proto_pois.trail(i));
+            }
+        }
+        for (int i = 0; i < output_message.category_size(); i++) {
+            remove_proto_child(output_message.mutable_category(i), category_includes);
```
AsherGlick, line 165: We should add some explanation why we doing the work to copy all the markers and then removing the ones we don't care about, instead of just only inserting the ones we do care about. Keep in mind doing this means that we have to spent the time to copy all the data, then each time we remove an element from an array we have to copy all the elements that are stored in memory after it to compensate for the now missing element.

AsherGlick (changes requested):

xml_converter/src/xml_converter.cpp
```diff
@@ -61,30 +63,115 @@ void write_xml_file(string xml_filepath, map<string, Category>* marker_categorie
...
 
     for (const auto& parsed_poi : *parsed_pois) {
         if (parsed_poi->classname() == "POI") {
             Icon* icon = dynamic_cast<Icon*>(parsed_poi);
             waypoint::Icon poi = icon->as_protobuf();
-            proto_message.add_icon()->CopyFrom(poi);
+            proto_pois.add_icon()->CopyFrom(poi);
+            map_ids.insert(icon->map_id);
         }
         if (parsed_poi->classname() == "Trail") {
```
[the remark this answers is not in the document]
klingbolt: I see why. Making that change

xml_converter/src/xml_converter.cpp
klingbolt: Good point. I adjusted the set `category_includes` So it now captures parents names. From your example, the set would now include 

```
mypack
mypack.easypath
mypack.easypath.trail
```

xml_converter/src/xml_converter.cpp
klingbolt: That makes logical sense to me. Declare when they need to be used.

xml_converter/src/xml_converter.cpp
```diff
@@ -61,30 +63,115 @@ void write_xml_file(string xml_filepath, map<string, Category>* marker_categorie
...
 }
 
+void create_proto_categories(waypoint::Category proto_category, set<string>* proto_categories) {
+    proto_categories->insert(proto_category.name());
+    for (int i = 0; i < proto_category.children_size(); i++) {
+        create_proto_categories(proto_category.children(i), proto_categories);
+    }
+}
+
+void remove_proto_child(waypoint::Category* proto_category, set<string> category_includes) {
```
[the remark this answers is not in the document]
klingbolt: I adjusted the set `category_includes` So it now captures the hierarchy of parents names. From your example, the set would now include

```
mypack
mypack.hardpath
mypack.hardpath.trail
mypack.trail
mypack.trail.hardpath
```

xml_converter/src/xml_converter.cpp
AsherGlick: The comment you added should really have a "we are doing a removal instead of an insertion because ...". This needs to be there because of the redundancy of "add everything then remove it" instead of the much simpler "only add what we need". The extra cost needs to be clearly  justifiable.

klingbolt: How does this sound instead 

```        
        // Marker_Categories has a tree hierarchy while POIS have a flat hierarchy.
        // We are doing a removal instead of an insertion because each parent category contains the data for all of its children.
        // It would be impractical to include all of the data from each category except for the children and then re-add the children. 
        // That would require coping every non-child attribute individually and iterating over all the children.
        // This pruning method is slower but ensures that the all wanted information is kept.
```

AsherGlick: Definitely better and makes clear the issue of not being able to "copy everything but".

However now it raises a new question which is: Why is the data formatted like that to begin with if its intention is to be serialized here?

AsherGlick: Keep in mind this is not me telling you that the entire structure of the code needs to be uprooted, it just needs to be explained here, even if the explanation is "we should uproot the entire structure of the code, but we are not going to do that right now". All that is needed is a logical thought through blurb so the next developer who reads it can understand why it is the way it is. For example it might be "we need to copy all the data be cause protobuf has a memory corruption issue if we try to share even a single string", this is not the case of course but a fresh pair of eyes wont know that right away and might just assume the code should up refactored only to run into said string corruption issue down the line. But if there is a comment saying "the code should be restructured in the future but we are doing this now because it should still be 'fast enough'" then the future developer knows that the reason it is being done in a weird way can be resolved with said refactor.

xml_converter/src/xml_converter.cpp
```diff
@@ -64,83 +65,85 @@ void write_xml_file(string xml_filepath, map<string, Category>* marker_categorie
...
+                vector<string> split_categories = split(proto_pois.icon(i).category().name(), ".");
+                auto category = proto_categories.find(lowercase(split_categories[split_categories.size() - 1]));
                 if (category == proto_categories.end()) {
                     cout << "Unknown MarkerCategory " << proto_pois.icon(i).category().name() << endl;
                 }
                 else {
-                    for (string split_category : split_categories) {
-                        category_includes.insert(split_category);
+                    // This loop adds the name of the category and all of it's parents
+                    string name;
```
AsherGlick, line 142: Is this string name just manually reassembling the name that was manually dissasembled earlier?
IE:
is `string name` equivalent to `proto_pois.icon(i).category().name()`  after going through this for loop?

xml_converter/src/xml_converter.cpp
```diff
@@ -149,20 +152,30 @@ void write_protobuf_file(string proto_filepath, map<string, Category>* marker_ca
...
                 }
                 output_message.add_trail()->CopyFrom(proto_pois.trail(i));
             }
         }
+        // In XML, Marker_Categories has a tree hierarchy while POIS have a flat hierarchy.
+        // This is preserved in the protobuf for ease of translation.
+        // We are doing a removal instead of an insertion because each parent category contains the data for all of its children.
+        // It would be impractical to include all of the data from each category except for the children and then re-add the children.
+        // That would require coping every non-child attribute individually and iterating over all the children.
+        // This pruning method is slower but ensures that the all wanted information is kept.
```
AsherGlick, line 176: Nit: This should be styled slightly better to not exceed the 80 character column mark. There are some unclear design patterns as to when to and when not to that I can only describe as "gut feeling", but plaintext English sentences at this indentation level fall into the category of "should be line wrapped at column 80".

Also this may be the first last and only time I find a spelling mistake but "This would require coping every..." should probably be "This would require copying every..."

xml_converter/src/xml_converter.cpp
```diff
@@ -64,83 +65,85 @@ void write_xml_file(string xml_filepath, map<string, Category>* marker_categorie
...
-                        category_includes.insert(split_category);
+                    // This loop adds the name of the category and all of it's parents
+                    string name;
+                    for (unsigned int j = 0; j < split_categories.size(); j++) {
+                        name += split_categories[j];
+                        category_includes.insert(name);
+                        name += ".";
                     }
                 }
                 output_message.add_icon()->CopyFrom(proto_pois.icon(i));
```
AsherGlick, line 149: we seem to call `add_icon()` on `proto_pois` only to then call `add_icon()` to copy the data out of `proto_pois`. Is there a reason we need `proto_pois`. This goes for the other `waypoint::Waypoint` variables that each hold a different element. This seems wasteful on the surface because all the data is being copied into a `waypoint::Waypoint` only to then be copied into another `waypoint::Waypoint` slightly later. While I believe the pruning argument for "copy and delete" holds water this "copy and then copy again" needs a robust case to be made as to why the data is not just copied once into the final location.

AsherGlick (changes requested):

xml_converter/src/xml_converter.cpp
```diff
@@ -64,83 +65,85 @@ void write_xml_file(string xml_filepath, map<string, Category>* marker_categorie
 }
 
 void create_proto_categories(waypoint::Category proto_category, set<string>* proto_categories) {
```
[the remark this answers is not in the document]
klingbolt: Added comments for both the create and remove functions to show what I expect the data in each set to be.

xml_converter/src/xml_converter.cpp
klingbolt: Yes. By the end of the loop they are equivalent. The name is in the format `a.b.c.d` and I want to make the set include 
```
{
     a
     a.b
     a.b.c
     a.b.c.d
} 
```

AsherGlick: Yep that makes sense.

xml_converter/src/xml_converter.cpp
```diff
@@ -64,13 +64,38 @@ void write_xml_file(string xml_filepath, map<string, Category>* marker_categorie
     outfile.close();
 }
 
+////////////////////////////////////////////////////////////////////////////////
+// create_proto_categories
+//
+// Fills a set of the names of categories for validation purposes
```
AsherGlick, line 70: Be more descriptive here to avoid confusion over which set. Additionally the purpose of the set created is not mandated by this function in any way so that does not need to be included.

This function might also be better served with a new name to fit its current function, as it is not actually creating a waypoint::Category protobuf object and `proto_categories` is just the name of an argument.

xml_converter/src/xml_converter.cpp
```diff
@@ -64,13 +64,38 @@ void write_xml_file(string xml_filepath, map<string, Category>* marker_categorie
...
     proto_categories->insert(lowercase(proto_category.name()));
     for (int i = 0; i < proto_category.children_size(); i++) {
         create_proto_categories(proto_category.children(i), proto_categories);
     }
 }
 
+////////////////////////////////////////////////////////////////////////////////
+// remove_proto_child
+//
+// Compares the relative name of the category against a set of the names of POIS
```
AsherGlick, line 89: Again be a bit more descriptive here about what your example is for.

xml_converter/src/xml_converter.cpp
```diff
@@ -96,32 +121,26 @@ void write_protobuf_file(string proto_filepath, map<string, Category>* marker_ca
...
     waypoint::Waypoint all_categories;
     std::set<string> proto_categories;
     for (const auto& category : *marker_categories) {
         waypoint::Category proto_category = category.second.as_protobuf();
         all_categories.add_category()->CopyFrom(proto_category);
         create_proto_categories(proto_category, &proto_categories);
     }
-    // Converts vector<Parseable*> to Waypoint::Icon and Waypoint::Trail
-    waypoint::Waypoint proto_pois;
+    // Collects a set of map ids from vector<Parseable*>
```
AsherGlick, line 132: `vector<Parseable*>` is a type, you are actually collecting map ids from `parsed_pois`, but you can do better then just saying "from parsed_pois" and instead include "from [[English description of what parsed pois is]]".

EG: An orange juicer note could say "Removes the pulp from the contents of harvester 6" or "Removes the pulp from the oranges" (the oranges which happen to be in harvester 6).

xml_converter/src/xml_converter.cpp
```diff
@@ -130,50 +149,58 @@ void write_protobuf_file(string proto_filepath, map<string, Category>* marker_ca
...
         // This is preserved in the protobuf for ease of translation.
-        // We are doing a removal instead of an insertion because each parent category contains the data for all of its children.
-        // It would be impractical to include all of the data from each category except for the children and then re-add the children.
-        // That would require coping every non-child attribute individually and iterating over all the children.
-        // This pruning method is slower but ensures that the all wanted information is kept.
+        // We are doing a removal instead of an insertion because each parent
+        // category contains the data for all of its children.
+        // It would be impractical to include all of the data from each category
+        // except for the children and then iterating over all the children.
+        // This pruning method is slower but ensures that information is kept.
```
AsherGlick, line 203:
```suggestion
        // In the XML, MarkerCategories have a tree hierarchy while POIS have a
        // flat hierarchy. This is preserved in the protobuf for ease of
        // translation. We are doing a removal instead of an insertion because
        // each parent category contains the data for all of its children. It
        // would be impractical to include all of the data from each category
        // except for the children and then iterating over all the children.
        // This pruning method is slower but ensures that information is kept.
```

xml_converter/src/xml_converter.cpp
```diff
@@ -130,50 +149,58 @@ void write_protobuf_file(string proto_filepath, map<string, Category>* marker_ca
...
                     }
+                    output_message.add_trail()->CopyFrom(proto_trail);
                 }
-                output_message.add_trail()->CopyFrom(proto_pois.trail(i));
             }
         }
-        // In XML, Marker_Categories has a tree hierarchy while POIS have a flat hierarchy.
+        // In the XML, Marker_Categories is a tree hierarchy while
+        // POIS have a flat hierarchy.
         // This is preserved in the protobuf for ease of translation.
```
AsherGlick, line 198:
```suggestion
```

xml_converter/src/xml_converter.cpp
AsherGlick: Provided a suggustion.

xml_converter/src/xml_converter.cpp
```diff
@@ -130,50 +149,58 @@ void write_protobuf_file(string proto_filepath, map<string, Category>* marker_ca
...
+                        cout << "Unknown MarkerCategory " << proto_trail.category().name() << endl;
+                    }
+                    else {
+                        // This loop adds the name of the category and all of it's parents
+                        string name;
+                        for (unsigned int j = 0; j < split_categories.size(); j++) {
+                            name += split_categories[j];
+                            category_includes.insert(name);
+                            name += ".";
+                        }
```
AsherGlick, line 190: This is a lot of duplicated logic, maybe abstract this to a function that takes in a `map_id` integer and a `category` string?

xml_converter/src/xml_converter.cpp
```diff
@@ -63,106 +64,145 @@ void write_xml_file(string xml_filepath, map<string, Category>* marker_categorie
...
-                else {
-                    for (string split_category : split_categories) {
-                        category_includes.insert(split_category);
+        for (const auto& parsed_poi : *parsed_pois) {
+            if (parsed_poi->classname() == "POI") {
+                Icon* icon = dynamic_cast<Icon*>(parsed_poi);
+                if (icon->map_id == map_id) {
+                    waypoint::Icon proto_icon = icon->as_protobuf();
+                    vector<string> split_categories = split(proto_icon.category().name(), ".");
+                    auto category = proto_categories.find(lowercase(split_categories[split_categories.size() - 1]));
```
AsherGlick, line 158: See earlier comment about this logic.

xml_converter/src/xml_converter.cpp
```diff
@@ -63,106 +64,145 @@ void write_xml_file(string xml_filepath, map<string, Category>* marker_categorie
...
-                }
-                else {
-                    for (string split_category : split_categories) {
-                        category_includes.insert(split_category);
+            else if (parsed_poi->classname() == "Trail") {
+                Trail* trail = dynamic_cast<Trail*>(parsed_poi);
+                if (trail->map_id == map_id) {
+                    waypoint::Trail proto_trail = trail->as_protobuf();
+                    vector<string> split_categories = split(proto_trail.category().name(), ".");
+                    auto category = proto_categories.find(lowercase(split_categories[split_categories.size() - 1]));
```
AsherGlick, line 179: See earlier comment about this logic.

AsherGlick (changes requested):

xml_converter/src/xml_converter.cpp
```diff
@@ -130,50 +149,58 @@ void write_protobuf_file(string proto_filepath, map<string, Category>* marker_ca
...
+                    else {
+                        // This loop adds the name of the category and all of it's parents
+                        string name;
+                        for (unsigned int j = 0; j < split_categories.size(); j++) {
+                            name += split_categories[j];
+                            category_includes.insert(name);
+                            name += ".";
+                        }
                     }
+                    output_message.add_icon()->CopyFrom(proto_icon);
```
AsherGlick, line 171: Does this need to be a `CopyFrom()`? What is the memory ownership of the data?

xml_converter/src/xml_converter.cpp
```diff
@@ -130,50 +149,58 @@ void write_protobuf_file(string proto_filepath, map<string, Category>* marker_ca
...
+                    else {
+                        // This loop adds the name of the category and all of it's parents
+                        string name;
+                        for (unsigned int j = 0; j < split_categories.size(); j++) {
+                            name += split_categories[j];
+                            category_includes.insert(name);
+                            name += ".";
+                        }
                     }
+                    output_message.add_trail()->CopyFrom(proto_trail);
```
AsherGlick, line 192: Related to the POI, does this need to be `CopyFrom`? Who owns the data?

xml_converter/src/xml_converter.cpp
klingbolt: I reworded the functions so that the POIs now search for the entire name e.g. `mypack.hardpath.trail`

Whenever the name is used in the program it should now be in that form.

xml_converter/src/xml_converter.cpp
klingbolt: From my understanding of the documentation says that `CopyFrom()` and `MergeFrom()` are my best options for setting a repeated message. `Copyfrom()`: "The method clears the current message and then merges the specified message using MergeFrom." and `MergeFrom()` "This method merges the contents of the specified message into the current message. Singular fields that are set in the specified message overwrite the corresponding fields in the current message. Repeated fields are appended. Singular sub-messages and groups are recursively merged." 
There is no function such as `set_icon()` available. 

So in these instances, I think `MergeFrom()` is more appropriate since `add_icon()` returns a pointer to an empty message and thus there is no data to clear. 
The message itself is actually created in the `as_protobuf()` function. I have changed the code so that the output of this function goes directly into `out_message` instead of a separate variable.

xml_converter/src/xml_converter.cpp
```diff
@@ -121,15 +115,13 @@ void write_protobuf_file(string proto_filepath, map<string, Category>* marker_ca
         cout << "Error making ./protobins" << endl;
         throw std::error_code();
     }
-    // Converts map<string, Category> to a Waypoint::Category
+    // Creates a Waypoint message that contains all categories
```
AsherGlick, line 118: This does not quite describe the next line, but more the next block of code. Maybe change this to:

"Creates a Waypoint message that will be filled with all categories"

AsherGlick (approved):

PR #120, opened 2023-01-20
improved error handling code in taco converter script

ceeb9:
rewrote the error handling logic. it now takes an enum representing the level of severity to determine what to do with the error. also added a try statement around parsing the xml files.

### DIFF
--- a/scripts/taco_to_burrito.py
+++ b/scripts/taco_to_burrito.py
@@ -7,6 +7,7 @@ from typing import *
 import json
 import sys
 from enum import Enum
+from tabulate import tabulate
 
 output_dir = ""
 
@@ -17,7 +18,10 @@ def main():
     if len(sys.argv) != 3:
         print(f"USAGE: {sys.argv[0]} SRC_DIR DEST_DIR")
         sys.exit(1)
-
+        
+    global error_in_new_file
+    global files_with_errors
+    global error_total
     global source_path
     global target_path
 
@@ -30,33 +34,51 @@ def main():
     for file in os.scandir(source_path):
         if(file.name.endswith(".xml")):
             full_path: str = os.path.join(source_path, file.name)
+            error_in_new_file = True
             convert_markers(full_path, target_path)
+            
+    print_error_stats()
 
 ################################################################################
 # Quick error printing logic
 ################################################################################
 error_location = ""
+error_in_new_file: bool = True
+error_total = 0
+files_with_errors = list()
+
 class ErrorType(Enum):
     INFO = 0
     WARNING = 1
     ERROR = 2
     
-def eprint(*args):
-    error_level = ErrorType.WARNING
-    for arg in list(args):
-        if(type(arg) == ErrorType):
-            error_level = arg
-            
-    print(error_level.name + " in file \"" + error_location + "\": ")
+def eprint(*args, error_level = ErrorType.WARNING):
+    global error_in_new_file
+    global error_total
+    error_total += 1
+    
+    if error_in_new_file:
+        print(error_level.name + " in file \"" + error_location + "\": ")
+        files_with_errors.append(error_location)
+        error_in_new_file = False
+
     for arg in list(args):
         if(type(arg) != ErrorType):
             print("    " + str(arg))
-    print()
+    print("    Skipping file...", end="\n\n")
     
     if(error_level == ErrorType.ERROR):
         print("Fatal error, exiting.")
         exit()
-            
+
+def print_error_stats():
+    global error_total
+    global files_with_errors
+    print("Conversion finished, found " + str(error_total) + " errors in " + str(len(files_with_errors)) + " files:\n    ")
+    for file in files_with_errors:
+        print(str(file))
+    
+        
 ################################################################################
 #
 ################################################################################
@@ -114,7 +136,6 @@ def parse_marker_category(marker_category_node, base=""):
     if "texture" in attribs:
         limited_attribs["texture"] = attribs["texture"]
 
-
     name = attribs["name"]
 
     metadata_tree[base + name] = limited_attribs
@@ -131,12 +152,13 @@ def parse_marker_category(marker_category_node, base=""):
 ################################################################################
 def convert_markers(xml_path: str, output_dir: str):
     global error_location
+    
     error_location = str(xml_path)
+    
     try:
         tree = ET.parse(xml_path)
     except Exception as e:
-        eprint(ErrorType.ERROR, "Error trying to parse XML, likely incorrectly formatted:", str(e))
-        
+        eprint("Error trying to parse XML, likely incorrectly formatted:", str(e), error_level=ErrorType.ERROR)
     else:
         root = tree.getroot()
     
@@ -151,7 +173,7 @@ def convert_markers(xml_path: str, output_dir: str):
         eprint("Root has {} children instead of the expected 2".format(len(root)))
         return
 
-    if root[0].tag != "MarkerCategory":
+    if root[0].tag != "MarkerCatego6ry":
         eprint("First element of root is {} not MarkerCategory".format(root[0].tag))
     else:
         marker_metadata = parse_marker_category(root[0])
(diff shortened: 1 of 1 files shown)


scripts/taco_to_burrito.py
```diff
@@ -32,25 +33,30 @@ def main():
...
-    print("   ", *args, **kwargs)
-
-
+error_location = ""
+class ErrorType(Enum):
+    INFO = 0
+    WARNING = 1
+    ERROR = 2
+    
+def eprint(*args):
```
AsherGlick, line 44: You can probably change this to
```python
def eprint(error_level, *args, **kwargs)
```
To get rid of the `for arg in list(args):` loops. If you want to assign a default value to it there might be some additional changes I would have to experiment with to find out, eg: it might require you to do `def eprint(*args, error_level=ErrorType.WARNING, **kwargs)` instead.

scripts/taco_to_burrito.py
```diff
@@ -32,25 +33,30 @@ def main():
...
+    error_level = ErrorType.WARNING
+    for arg in list(args):
+        if(type(arg) == ErrorType):
+            error_level = arg
+            
+    print(error_level.name + " in file \"" + error_location + "\": ")
```
AsherGlick, line 50: This is going to print out the file name for every single error. This ended up being a ton of unnecessary extra noise in the output. The previous version had logic to prevent this extra noise and only print out the file when the first error in that file was printed, that functionality should be preserved.

scripts/taco_to_burrito.py
```diff
@@ -32,25 +33,30 @@ def main():
...
+    for arg in list(args):
```
AsherGlick, line 51: This looks like it is going to print each argument on a separate indented line.
```python
eprint("Hello", "World")
```
would print out
```
    Hello
    World
```
when it probably should print out
```
    Hello World
```

This can be fixed if you change the function signature mentioned previously and just call
```python
print("    ", *args, **kwargs)
```
like the previous function did.

scripts/taco_to_burrito.py
```diff
@@ -125,11 +130,17 @@ def parse_marker_category(marker_category_node, base=""):
...
-    # print(xml_path)
-    tree = ET.parse(xml_path)
-    root = tree.getroot()
-
+    global error_location
+    error_location = str(xml_path)
+    try:
+        tree = ET.parse(xml_path)
+    except Exception as e:
+        eprint(ErrorType.ERROR, "Error trying to parse XML, likely incorrectly formatted:", str(e))
```
AsherGlick, line 138: Generally you dont want optional positional arguments to be present first in a function signature. If this is changed into a keyword argument then it should become:
```python
eprint("Error trying to parse XML, likely incorrectly formatted:", str(e), error_type=ErrorType.ERROR)
```

AsherGlick (changes requested):
It is always great to see new contributors. Overall I like the idea of passing an enum, and wrapping the xml parser in a try catch block is definitely a good idea.

I did not get around to running this code so most of my notes are just guesses and might be wrong. Feel free to let me know if my assumptions are incorrect.

scripts/taco_to_burrito.py
ceeb9: i was getting some weird behavior earlier when i was trying this approach instead. seems like moving the error_level argument to after *args solves it, and lets me set a default value.

scripts/taco_to_burrito.py
ceeb9: this is intended, so that you can pass an explanation and the error itself as two different arguments, like this for example:
image
the only thing that changes is that you have to do a little bit of string formatting when calling the eprint function which is already being done anyway.

scripts/taco_to_burrito.py
ceeb9: in practice that doesn't happen: as far as i can see, the only time eprint is called is on the root node of the XML file. in the case of there being malformed XML, sure it might happen more than once in a file, but by that time the script has exited. i can't replicate a situation where there are multiple errors in one file, but i'll add a quick bool to check for this.

scripts/taco_to_burrito.py
ceeb9: locally this is what i have... guess i changed it after the commit or something

scripts/taco_to_burrito.py
```diff
@@ -7,6 +7,7 @@
 import json
 import sys
 from enum import Enum
+from tabulate import tabulate
```
AsherGlick, line 10: This is not used anywhere

scripts/taco_to_burrito.py
```diff
@@ -151,7 +173,7 @@ def convert_markers(xml_path: str, output_dir: str):
         eprint("Root has {} children instead of the expected 2".format(len(root)))
         return
 
-    if root[0].tag != "MarkerCategory":
+    if root[0].tag != "MarkerCatego6ry":
```
AsherGlick, line 176: This has a 6 in it. In the future, please make sure simple issues like this are corrected before asking for a review.

scripts/taco_to_burrito.py
```diff
@@ -30,33 +34,51 @@ def main():
     for file in os.scandir(source_path):
         if(file.name.endswith(".xml")):
             full_path: str = os.path.join(source_path, file.name)
+            error_in_new_file = True
```
AsherGlick, line 37: Instead of having several uncorrelated global flags that must be set in order for eprint to work properly just have a single setter that sets all of the values at once so future users are less likely to accidentally break something in a non obvious way..

scripts/taco_to_burrito.py
```diff
@@ -30,33 +34,51 @@ def main():
...
+    if error_in_new_file:
+        print(error_level.name + " in file \"" + error_location + "\": ")
+        files_with_errors.append(error_location)
+        error_in_new_file = False
+
     for arg in list(args):
         if(type(arg) != ErrorType):
             print("    " + str(arg))
-    print()
+    print("    Skipping file...", end="\n\n")
```
AsherGlick, line 68: Remove this print it is printing two lines for every `eprint()` call and not providing any useful information.

scripts/taco_to_burrito.py
```diff
@@ -30,33 +34,51 @@ def main():
...
         print("Fatal error, exiting.")
         exit()
-            
+
+def print_error_stats():
+    global error_total
+    global files_with_errors
+    print("Conversion finished, found " + str(error_total) + " errors in " + str(len(files_with_errors)) + " files:\n    ")
+    for file in files_with_errors:
+        print(str(file))
```
AsherGlick, line 79: Don't print out all the file names at the end, this is more clutter that requires the user to scroll up to see any actual information.

scripts/taco_to_burrito.py
```diff
@@ -30,33 +34,51 @@ def main():
...
     
     if(error_level == ErrorType.ERROR):
         print("Fatal error, exiting.")
         exit()
-            
+
+def print_error_stats():
+    global error_total
+    global files_with_errors
+    print("Conversion finished, found " + str(error_total) + " errors in " + str(len(files_with_errors)) + " files:\n    ")
```
AsherGlick, line 77: If you are explicitly adding the functionality to differentiate between errors and warnings then they should be separated when printing out the totals.

scripts/taco_to_burrito.py
AsherGlick: I don't see a solid argument for changing the normal functionality and removing all the keyword functionality in favor of mandating a newline per argument. If you want a newline just do the formatting to add the newline instead of require formatting to not add it.

scripts/taco_to_burrito.py
```diff
@@ -30,33 +34,51 @@ def main():
...
-            error_level = arg
-            
-    print(error_level.name + " in file \"" + error_location + "\": ")
+def eprint(*args, error_level = ErrorType.WARNING):
+    global error_in_new_file
+    global error_total
+    error_total += 1
+    
+    if error_in_new_file:
+        print(error_level.name + " in file \"" + error_location + "\": ")
```
AsherGlick, line 61: Now that you are only printing out this line once per file you are only getting the error level of the first error message. If a warning is printed followed by an info or an error then the fact that the other messages are info or error will be hidden.

AsherGlick (changes requested):

PR #127, opened 2023-02-27
Allow Sane Indendentation in Jinja Templates

AsherGlick:
A simple Jinja2 plugin was written that acts as a preprocessor and aligns the indents of all of the blocks to the indentation level of the block start. This allows for very rudimentary indentation within the jinja template without that indentation carrying over to the generated files.

This PR will likely need to be rebased against the other PRs currently waiting to be merged into `xml_converter`, it will be easier to just reapply the indentation changes to the templates in this PR then it will be to apply these changes to other PRs because we can validate in this PR that none of the generated files themselves have changed. Therefore this should be blocked on #114 and #124 being merged first.

This extension may eventually become its own project, and this is just a first draft to allow for way easier code reviews on the template files themselves. Because such, documenting this extension is not a priority.

### DIFF
--- a/xml_converter/generators/code_generator.py
+++ b/xml_converter/generators/code_generator.py
@@ -7,6 +7,7 @@ import os
 import markdown
 from dataclasses import dataclass, field
 from jinja2 import Template, FileSystemLoader, Environment
+from jinja_helpers import UnindentBlocks
 
 SchemaType = Dict[str, Any]
 schema = """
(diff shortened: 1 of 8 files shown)


klingbolt (approved):
Works as intended and makes the template files much easier to understand

AsherGlick:
Rebased changes to make the diff of this pr easier to verify, by having it once again be just whitespece changes for the effected code.

PR #131, opened 2023-03-05
Removing artifact retention 7 day limit

AsherGlick:
In theory the 7 day limit seemed economical, but in practice not having easy access to whatever the "newest" build artifact on master is has proven annoying.


klingbolt:
How long is default retention?

Issue #77, opened 2022-07-24
[XML Markers] Tree menu GUI for Marker Category Selection


AsherGlick:
Once marker categories can be parsed by Burrito we will now be able to know what categories we can filter on.

- By default all markers categories will be off
- When a marker category is turned on, it will be turned on for the entire world, even though the protobin files are split up per map.
- If a category is on but there is no protobin for that category in the current map the category will stay on for whenever the player enters a map that does have the category
 - The user will be able to open and close the tree-view the same way they are currently able to save and load markers
- One category will always be the "selected" category even though multiple may be visible. Only this category can be edited. To edit a different category the user must "select" a different category
- Ideally the user selection of which categories they have on or off will be persisted between application restarts

Godot has a built-in tree view, though it is only really controllable via code and does not have a great manual UI in the Godot editor. [Godot Docs](https://docs.godotengine.org/en/stable/classes/class_tree.html)

PR #132, opened 2023-03-12
Godot category selection

klingbolt:
Dependent on PR #124 

Addressing issue #77


Spatial.tscn
```diff
@@ -483,9 +484,9 @@ window_title = "Open a File"
 mode = 0
 access = 2
 filters = PoolStringArray( "*.png" )
-current_dir = "."
+current_dir = "/home/steph/Code/Projects/Burrito"
```
klingbolt, line 487: I'm not sure let why this keeps changing without my knowledge.

AsherGlick: This is why I always do a `git diff <filename>` before `git add <filename>` and always do each file individually instead of `git add *`. You always want to know what you are adding.

This particular issue is a weird bug, idk if it is fixed yet or not but when you brows to a particular location it changes this path. The top-level `presubmit.sh` has 3 commands that work to remove all instances of it.

klingbolt:
Closing this PR. New PR is #141

Issue #78, opened 2022-07-24
[XML Markers] Save Marker Protobins from GDScript


AsherGlick:
Currently there is a way to save marker files after they have been edited in-engine. However this only works because you can currently only have one `.json` file of markers open, and when you save you just save that one file.

Now that there are multiple files we must rely on the "selected" category for editing a category.

However when a user wants to save their changes there are a few questions, should ALL changes be saved?
Even ones that are not in the currently selected category?
Additionally should changes to markers not on the current map be saved?
If there are changes on the current map and the player transitions, should we automatically save so that when the player returns their changes are persisted?

The saved protobin files will clearly be able to be opened again by the user, but eventually they will also be able to be exported so that other users may access them too. So the answers to the decisions above must be correct for both an end user and a marker pack designer.

klingbolt:
I think a potential solution for this is to have a separate folder that contains local data when any changes are made. i.e.

```xml_converter/protobins/*.data
xml_converter/local_protobins/*.data```

We could have burrito load up `local_protobins/*.data` first if it exists and then load `protobins/*.data` and skip any paths or icons that have the same GUID. This would allow the user's edits to persist between maps and between sessions without the pack data being changed. 

For saving, I think there should be a specific dialog to confirm that you want to overwrite the file. I think this should be done so that saves are intentional. So I'm thinking of a dialog for `Local Edits` that shows all the categories that have been edited in the current map. The user would then select a category and press a button on whether to save the data to the main file or delete the local data.

AsherGlick:
Something irks me about using GUID to skip paths, idk exactly what yet. Can treat the `local_protobins` folder as a "recovery" or autosave folder. So if files in that folder exist we ask "do you want to load the previous unsaved markers?" If they click yes we just load those protobins instead of the normal ones? Maybe some sort of filtering still needs to exist for like if you edit one marker pack you might not want other marker packs to be edited (aka reverted) when you load your previously saved marker packs after upgrading a different external pack? Maybe the filtering is on marker category instead of GUID? We could also implement an autosave feature where x seconds after your last change it saves to `local_protobins` unless there is another change within that time.

klingbolt:
I found a flaw with using GUID. If they deleted a path or icon, we wouldn't have a way of knowing and the next load would bring it back.

klingbolt:
[...]

PR #154, opened 2023-08-24
 [XML Markers] Save Protobins from GDScript

klingbolt:
Addressing Issue  [XML Markers] Save Marker Protobins from GDScript #78 

Addresses the questions posed in the issue by saving all data when the save button is pressed. The average user would not make a distinction of saving one pack's data and not another. Same principle applies to switching maps. By saving all data, you are also not required to be in the map where the changes have been made.  

Finally, an autosave check was added before a new maps data is loaded. This could be augmented further to catch changes within x amount of time.

### DIFF
--- a/Gizmo/PointEdit.gd
+++ b/Gizmo/PointEdit.gd
@@ -47,9 +47,11 @@ func update_point():
 	if self.translation != self.last_translation:
 		if point_type == "path" || point_type == "area":
 			self.object_link.set_point_position(self.object_index, self.translation)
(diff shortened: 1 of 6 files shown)


AsherGlick:
Split this up into 2 PRs. One for saving and one for the second half of the category being a parent of icons and trails in gdscript.

klingbolt:
https://github.com/AsherGlick/Burrito/pull/157

klingbolt:
Split into #157 and #158

PR #157, opened 2023-08-25
[XML Converter] Reading waypoint

klingbolt:
Split from  #154

Contains only the new reading functionality from the change to Waypoint files over Json


Spatial.gd
```diff
@@ -284,27 +285,34 @@ func reset_minimap_masks():
 		minimap_path.material.set_shader_param("minimap_corner", compass_corner1)
 		minimap_path.material.set_shader_param("minimap_corner2", compass_corner2)
 
-var markerdata = Waypoint.Waypoint.new()
-var marker_file_dir = "user://protobins/"
+var Waypoint_data = Waypoint.Waypoint.new()
+var marker_file_dir = Directory.new()
```
AsherGlick, line 289: Now that this is being initialized in the _ready() function we should declare it before the _ready() function or figure out some way to set this value here with an onready like trigger

Spatial.gd
```diff
@@ -284,27 +285,34 @@ func reset_minimap_masks():
...
 var marker_file_path = ""
-var marker_packs_array = Array()
+var root: TreeItem
+
+##########Node Connections###########
 onready var marker_packs = $Control/Dialogs/MarkerPacks/MarkerPacks
+onready var icons = $Icons
+onready var paths = $Paths
+onready var minimap = $Control/MiniMap
```
AsherGlick, line 297: lets add types to these variables because they are basically constant references to the godot scene nodes.
Is `onready const` a thing? I feel like it is not but it might be.

Spatial.gd
```diff
@@ -284,27 +285,34 @@ func reset_minimap_masks():
...
 func load_waypoint_markers(map_id):
-	self.marker_file_path = self.marker_file_dir + String(map_id) + ".data"
-	print("Loading protobuf file from path ", self.marker_file_path)
-	self.markerdata.clear_category()
-	self.markerdata.clear_icon()
-	self.markerdata.clear_trail()
+	self.marker_file_path = self.marker_file_dir.get_current_dir() + String(map_id) + ".data"
+	self.Waypoint_data.clear_category()
+	clear_map_markers()
+	root.free()
```
AsherGlick, line 304: You probably don't need to do this because in the next function `init_category_tree()` just overwrites it and the old data will be garbage collected.

Spatial.gd
```diff
@@ -284,27 +285,34 @@ func reset_minimap_masks():
 		minimap_path.material.set_shader_param("minimap_corner", compass_corner1)
 		minimap_path.material.set_shader_param("minimap_corner2", compass_corner2)
 
-var markerdata = Waypoint.Waypoint.new()
-var marker_file_dir = "user://protobins/"
+var Waypoint_data = Waypoint.Waypoint.new()
```
AsherGlick, line 288:
```suggestion
var waypoint_data = Waypoint.Waypoint.new()
```

Spatial.gd
```diff
@@ -397,69 +401,20 @@ func gen_map_markers():
...
-		add_category(root, category, category.get_name(), false)
 
 
-func add_category(item: TreeItem, category, full_category_name: String, collapsed: bool):
+func parse_Waypoint():
+	for category in self.Waypoint_data.get_category():
+		parse_category(root, category, category.get_name(), false)
+
+
+func parse_category(item: TreeItem, category, full_category_name: String, collapsed: bool):
```
AsherGlick, line 417: This might want to be something like `create_nodes_from_waypoint_category`.

Spatial.gd
```diff
@@ -397,69 +401,20 @@ func gen_map_markers():
...
 	root.set_selectable(0, false)
 	root.set_text(1, "Visible")
-	
-	for category in self.markerdata.get_category():
-		self.marker_packs_array.append(category.get_name())
-		add_category(root, category, category.get_name(), false)
 
 
-func add_category(item: TreeItem, category, full_category_name: String, collapsed: bool):
+func parse_Waypoint():
```
AsherGlick, line 412: This might want to be something more descriptive to indicate that it is creating nodes from the waypoint category object.

Spatial.gd
```diff
@@ -586,39 +565,10 @@ func gen_new_icon(position: Vector3, texture_path: String, waypoint_icon, catego
...
-		for point in range(path.get_point_count()):
-			var point_position:Vector3 = path.get_point_position(point)
-			points.append([point_position.x, point_position.y, -point_position.z])
-		paths_data.append({
-			"points": points,
-			"texture": path.texture_path
-		})
-
-	var data_out = {"icons": icons_data, "paths": paths_data}
-	return data_out
```
AsherGlick, line 613: This should be a part of the next PR, no need to delete stuff with TODOs of re-adding it later if you dont need to delete it for the work you are currently doing to function.

Spatial.gd
```diff
@@ -771,21 +734,17 @@ func _on_NewPathPoint_pressed():
...
 
 ################################################################################
-# open the save dialog window. When a path is selected
-# _on_SaveDialog_file_selected() will be called with the user specified path.
+# 
 ################################################################################
 func _on_SavePath_pressed():
-	$Control/Dialogs/SaveDialog.show()
+	#TODO: Save to Waypoint
+	pass
```
AsherGlick, line 741: Revert this for the same save reasons as earlier.

Spatial.gd
```diff
@@ -771,21 +734,17 @@ func _on_NewPathPoint_pressed():
...
-# Save the current markers to a file, this includes all markers in memory not
-# just the markers on the current map.
+# TODO: This function will be used when exporting packs
 ################################################################################
 func _on_SaveDialog_file_selected(path):
-	self.markerdata[str(self.map_id)] = data_from_renderview()
-	var save_game = File.new()
-	save_game.open(path, File.WRITE)
-	save_game.store_string(JSON.print(self.markerdata))
+	pass
```
AsherGlick, line 747: Revert this for the same save reasons as earlier.

AsherGlick (changes requested):

Spatial.gd
klingbolt: It turns out that without clearing it, its just continually adding to the tree. I moved `root.free()` into the function `init_category_tree()` if it is necessary part of the reseting the tree
image

Spatial.gd
AsherGlick: This should have had the comment reverted too.

Spatial.gd
AsherGlick: This should have had the comment reverted too.

Spatial.gd
```diff
@@ -48,6 +48,20 @@ var taco_parser: TacoParser
...
 
+# Scenes used throughout this scene
+var route_scene = load("res://Route.tscn")
+var icon_scene = load("res://Icon.tscn")
+var path2d_scene = load("res://Route2D.tscn")
+var gizmo_scene = load("res://Gizmo/PointEdit.tscn")
+
+##########Node Connections###########
+onready var marker_packs := $Control/Dialogs/MarkerPacks/MarkerPacks as Tree
+onready var root := self.marker_packs.create_item() as TreeItem
```
AsherGlick, line 59: You probably don't need to use .create_item() here because all it will do is create something that will be deleted.
```suggestion
onready var root: TreeItem
```

Spatial.gd
AsherGlick: this was never re-added anywhere

AsherGlick (changes requested):

AsherGlick (approved):

PR #158, opened 2023-08-25
[XML Markers] Save Protobins within Godot

klingbolt:
Split from #154 

Dependent on #157

Addressing Issue [XML Markers] Save Marker Protobins from GDScript https://github.com/AsherGlick/Burrito/issues/78

Addresses the questions posed in the issue by saving all data when the save button is pressed. The average user would not make a distinction of saving one pack's data and not another. Same principle applies to switching maps. By saving all data, you are also not required to be in the map where the changes have been made.

Finally, an autosave check was added before a new maps data is loaded. This could be augmented further to catch changes within x amount of time.


klingbolt:
Wrong Branch

PR #162, opened 2023-09-21
Updating Entrypoint Args

AsherGlick:
This will better reflect the ideal capabilities of the xml converter and will allow us to think about and write proper tests as we get into the final phases of developing this module.

While these arguments are not completely final, some additional arguments are necessary for determining how files should be outputted, they are pretty close.

### DIFF
--- a/xml_converter/src/xml_converter.cpp
+++ b/xml_converter/src/xml_converter.cpp
@@ -431,11 +431,6 @@ void write_taco_directory(string directory, map<string, Category>* marker_catego
     write_xml_file(xml_filepath, marker_categories, parsed_pois);
 }
 
-void convert_taco_directory(string directory, string output_directory, map<string, Category>* marker_categories, vector<Parseable*>* parsed_pois) {
-    move_supplementary_files(directory, output_directory);
-    read_taco_directory(directory, marker_categories, parsed_pois);
-}
-
 ////////////////////////////////////////////////////////////////////////////////
 // process_data
 //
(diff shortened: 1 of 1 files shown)


xml_converter/src/xml_converter.cpp
```diff
@@ -431,49 +431,77 @@ void write_taco_directory(string directory, map<string, Category>* marker_catego
...
+    vector<string> output_waypoint_paths,
+
+    // This is a special output path used for burrito internal use that splits
+    // the waypoint protobins by map id.
+    string output_split_waypoint_dir) {
+    // All of the loaded pois and categories
     vector<Parseable*> parsed_pois;
     map<string, Category> marker_categories;
 
-    try {
```
klingbolt, line 453: Can you explain why this try is being removed? There are cases when we want the program to finish processing all of the folders when one of the folders gives us an error.

xml_converter/src/xml_converter.cpp
```diff
@@ -488,34 +516,61 @@ void process_data(
 // receive.
 //
 // Example usage
-//   ./xml_converter --input-path ../packs/marker_pack --output-path ../output_packs
-//   ./xml_converter --input-path ../packs/* --output-path ../output_packs
+//   ./xml_converter --input-taco-paths ../packs/marker_pack --output-split-waypoint-path ../output_packs
```
klingbolt, line 519: `--input-taco-paths` should be changed to match the if statement in line 535.

klingbolt (changes requested):

xml_converter/src/xml_converter.cpp
AsherGlick: Summarizing offline discussion to this point.
* This was no longer needed because the marker pack directory used to be hardcoded, meaning that all marker packs in the folder were always parsed. Now it is dynamic, meaning that a user who specified a marker pack that is causing errors would just need to remove the specific marker pack from the command arguments.
* This however still runs across an issue when a user is providing an folder such as `./markerpacks/*'. zsh will automatically expand these arguments but bash will not. Meaning that a bash user would be in a similar situation as they were before, except now instead of just removing a marker pack listing they would have to add all of the other marker packs manually.
* But, to a real user, not a developer. Is there any real utility in creating an output from a partial marker pack, or partial set of marker packs? In general we should not be throwing any errors that need to be caught if we can help it. And if an uncaught error reaches this point should we really create a partial markerpack output, split at a possibly arbitrary point in the processing? I would argue it is better to output nothing, alongside a loud error, then quietly output partially corrupted data.

xml_converter/src/xml_converter.cpp
AsherGlick: This is a valid critique. Overall it is a minor detail, I will fix it in a future pr instead of trying to rebase other commits for the sake of a comment.

xml_converter/src/xml_converter.cpp
klingbolt: I hadn't considered the partial data. Thank you for clarifying. The thing about bash vs zsh is frustrating but not something most people will run into.

klingbolt (approved):

PR #168, opened 2023-09-30
String Hierarchy Utility Functions

AsherGlick:
This adds some string hierarchy utility functions that allow for some minor optimizations when calling string hierarchy functions with a single path, by not requiring that a vector be created to house the path.

It also introduces sub_hierarchy functionality so that hierarchy don't need to be queried from the root, allowing for recursive traversal the category tree without requiring a full hierarchy path be constructed or maintained.

~Lastly it fixes up some incorrect linter checks that were mostly passing because the linter was looking at the existing code to try and determine what the rule should be. But now it is explicitly that pointer `*`s should be attached to the type, and reference `&`s should be attached to the variable.~

The linter change turned out to be widespread. It will be moved to its own pr.

### DIFF
--- a/xml_converter/src/string_hierarchy.cpp
+++ b/xml_converter/src/string_hierarchy.cpp
@@ -5,6 +5,24 @@
 ////////////////////////////////////////////////////////////////////////////////
 // in_hierarchy
 //
+// Returns if a particular node exists at the top level of the hirearchy.
+////////////////////////////////////////////////////////////////////////////////
+bool StringHierarchy::in_hierarchy(
+    const std::string &node) const {
+    if (this->all_children_included) {
+        return true;
+    }
+
+    auto iterator = this->children.find(node);
+    if (iterator == this->children.end()) {
+        return false;
+    }
+    return true;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// in_hierarchy
+//
 // Returns if the given path is in the hierarchy or not
 ////////////////////////////////////////////////////////////////////////////////
 bool StringHierarchy::in_hierarchy(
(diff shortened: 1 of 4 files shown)


xml_converter/src/string_hierarchy.cpp
```diff
@@ -12,6 +30,19 @@ bool StringHierarchy::in_hierarchy(
...
 }
 
+////////////////////////////////////////////////////////////////////////////////
+// in_hierarchy
+//
+// An explicit version of in_hierarchy that takes an initalizer list to prevent
+// ambiguity between the vector and string overloads of the function.
+////////////////////////////////////////////////////////////////////////////////
+bool StringHierarchy::in_hierarchy(
+    const std::initializer_list<std::string> input) const {
```
klingbolt, line 40: This could be pass by reference to match the other functions.

xml_converter/src/string_hierarchy.cpp
```diff
@@ -39,6 +70,57 @@ bool StringHierarchy::_in_hierarchy(
...
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// sub_hierarchy
+//
+// An explicit version of sub_hierarchy that takes an initalizer list to
+// prevent ambiguity between the vector and string overloads of the function.
+////////////////////////////////////////////////////////////////////////////////
+const StringHierarchy *StringHierarchy::sub_hierarchy(
+    const std::initializer_list<std::string> input) const {
```
klingbolt, line 98: This could be pass by reference to match the other functions.

xml_converter/src/string_hierarchy.hpp
```diff
@@ -7,6 +7,21 @@
 class StringHierarchy {
  public:
     bool in_hierarchy(
+        const std::string &node) const;
+
+    bool in_hierarchy(
+        const std::vector<std::string> &path) const;
+
+    bool in_hierarchy(
+        const std::initializer_list<std::string>) const;
```
klingbolt, line 16: This could be pass by reference to match the other functions.

xml_converter/src/string_hierarchy.hpp
```diff
@@ -7,6 +7,21 @@
...
+
+    const StringHierarchy *sub_hierarchy(
+        const std::string &node) const;
+
+    const StringHierarchy *sub_hierarchy(
+        const std::initializer_list<std::string> input) const;
```
klingbolt, line 22: This could be pass by reference to match the other functions.

klingbolt (changes requested):

klingbolt (approved):

PR #169, opened 2023-09-30
Splitting up the Github CI Workflow to Build in Parallel

AsherGlick:
Burrito is made up of several components:
* xml_converter -the new xml parser code
* godot ui - the primary user interface
* burrito_link - the connection between the ui and gw2
* burrito-fg - a gdnative library to talk to X11
* taco_parser - an xml parser that will be replaced with xml_converter

This change causes those to all be built in parallel and then assembled at the end. One kinda big hack that is used is that burrito-fg and taco_parser are mocked when compiling the godot ui. Then once everything is compiled they are seamlessly swapped in before the final packaging.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,30 +26,36 @@ on:
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   # This workflow contains a single job called "build"
-  Build-Linux:
+  Build-XMLConverter-Linux:
     # The type of runner that the job will run on
     runs-on: ubuntu-20.04
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout repository
+        uses: actions/checkout@v4
 
       - name: Create the Output Directory
         run: mkdir -v -p output
 
-      # `clang-9` must be installed here because of a weird unlisted dependency
-      # on some sort of file that clang-9 installs. Without it iwyu would
-      # complain about missing files and error out.
-      - name: Install include-what-you-use
-        run: |
-          sudo apt-get install iwyu
-          sudo apt-get install clang-9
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -26,61 +26,133 @@ on:
...
     # The type of runner that the job will run on
     runs-on: ubuntu-20.04
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout repository
+        uses: actions/checkout@v4
 
       - name: Create the Output Directory
```
klingbolt, line 38: This step can be removed since the build is no longer using the output folder.

.github/workflows/main.yml
```diff
@@ -26,61 +26,133 @@ on:
...
+          path: xml_converter/build/xml_converter
+          if-no-files-found: error 
+
+  Build-BurritoLink-Linux:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Create the Output Directory
```
klingbolt, line 94: This step can be removed since the build is no longer using the output folder.

.github/workflows/main.yml
```diff
@@ -26,61 +26,133 @@ on:
...
+          path: taco_parser/target/release/libgw2_taco_parser.so
+          if-no-files-found: error 
+
+  Build-BurritoUI-Linux:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Create the Output Directory
```
klingbolt, line 154: This step can be removed since the build is no longer using the output folder.

klingbolt (changes requested):

klingbolt (approved):

PR #170, opened 2023-09-30
Removing Compatability Field

AsherGlick:
This field will be replaced with another solution that is centered around each product. One file for each product and in the file all of the support will be listed. This will be a better experience then every product trying to edit a field across multiple files.

### DIFF
--- a/xml_converter/generators/code_generator.py
+++ b/xml_converter/generators/code_generator.py
@@ -17,8 +17,6 @@ shared_field_properties: Dict[str, DefType] = {
     "type": string_t(),
     "name": string_t(),
     "applies_to": array_t(enum_t(["Icon", "Trail", "Category"])),
-    # To Be Depricated
-    "compatability": array_t(enum_t(["BlishHUD", "Burrito", "TacO"])),
     "xml_fields": array_t(string_t(pattern="^[A-Za-z]+$")),
     "protobuf_field": string_t(pattern="^[a-z_.]+$"),
 }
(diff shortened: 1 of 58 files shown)


klingbolt (approved):

xml_converter/generators/code_generator.py
```diff
@@ -589,7 +582,7 @@ def write_webdocs(self, output_directory: str) -> None:
             for field_row in field_rows:
                 complete_field_row_list.append(field_row)
 
-            with open(os.path.join(output_directory, page + "_gen.html"), 'w') as f:
+            with open(os.path.join(output_directory, page + ".html"), 'w') as f:
```
klingbolt, line 585: What the rational for removing "_gen" from a generated file?

klingbolt, line 585: What the rational for removing "_gen" from a generated file?

xml_converter/generators/code_generator.py
```diff
@@ -208,7 +201,7 @@ class Generator:
     def delete_generated_docs(self, dir_path: str) -> None:
         for filepath in os.listdir(dir_path):
             filepath = os.path.join(dir_path, filepath)
-            if filepath.endswith("_gen.hpp") or filepath.endswith("_gen.cpp") or filepath.endswith("_gen.html"):
+            if filepath.endswith("_gen.hpp") or filepath.endswith("_gen.cpp") or filepath.endswith(".html"):
```
klingbolt, line 204: What the rational for removing "_gen" from a generated file?

klingbolt (changes requested):

xml_converter/generators/code_generator.py
AsherGlick: The _gen suffix was never tested with the html directory and broke all the hyperlinks in the documentation pages when it was added. This fixes them at the expense of the suffix. But the entire html structure is generated, there are no manual parts, so there is no need to delineate between manually created files and the generated files when deleting them as is necessary for the code files. This was shown to be sufficient back before the _gen suffix was added and the html documentation was the only thing being generated.

xml_converter/generators/code_generator.py
AsherGlick: Responded in the other comment to answer the same question.

klingbolt (approved):

PR #173, opened 2023-09-30
Creating Simpler Nested Proto Parse Templates

AsherGlick:
The previous logic had duplicated code chunks. The new logic is much more flexible and unifies the template branches to remove duplicate chunks.

### DIFF
--- a/xml_converter/generators/cpp_templates/class_template.cpp
+++ b/xml_converter/generators/cpp_templates/class_template.cpp
@@ -100,7 +100,7 @@ waypoint::{{cpp_class}} {{cpp_class}}::as_protobuf() const {
     {% endif %}
     {% for attribute_variable in attribute_variables %}
         {% if attribute_variable.is_component == false %}
-            {% if (attribute_variable.is_trigger == true)%}
+            {% if (attribute_variable.proto_drilldown_calls != "")%}{# TODO: This is a hack to preserve functionality when removing is_trigger #}
                 {% if (attribute_variable.attribute_type == "Custom")%}
                     if (this->{{attribute_variable.attribute_flag_name}}) {
                         if (trigger == nullptr) {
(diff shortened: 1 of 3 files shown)


xml_converter/generators/cpp_templates/class_template.cpp
```diff
@@ -149,55 +149,28 @@ waypoint::{{cpp_class}} {{cpp_class}}::as_protobuf() const {
...
-                        this->{{attribute_variable.attribute_name}} = proto_{{cpp_class_header}}.{{attribute_variable.protobuf_field}}();
-                        this->{{attribute_variable.attribute_flag_name}} = true;
-                    }
-                {% endif %}
+            {% if (attribute_variable.attribute_type in ["MultiflagValue", "CompoundValue", "Custom", "CompoundCustomClass"]) %}
+                if (proto_{{cpp_class_header}}{{attribute_variable.proto_drilldown_calls}}.has_{{attribute_variable.protobuf_field}}()) {
+                    this->{{attribute_variable.attribute_name}} = from_proto_{{attribute_variable.class_name}}(proto_{{cpp_class_header}}{{attribute_variable.proto_drilldown_calls}}.{{attribute_variable.protobuf_field}}());
+                    this->{{attribute_variable.attribute_flag_name}} = true;
+                }
+            {% elif (attribute_variable.class_name == "string") %}{# TODO: why is this .class_name when the others are .attribute_name #}
```
klingbolt, line 159: I think this TODO is meant to say `attribute_type` instead of name. I think it can be changed to `attribute_type` without issue since `class_name` is directly assigned by `attribute_type` in the generator for String types. Can be a separate PR

klingbolt (approved):

xml_converter/generators/cpp_templates/class_template.cpp
AsherGlick: Yes it is supposed to be `attribute_type`. It will changed in a later PR.

PR #178, opened 2023-10-09
First bool Xml test

klingbolt:
This set of tests is to build on the framework and test a value being copied and a value having an acceptable input but the output being different.


xml_converter/test_cases/can_fade_tests/can_fade_is_false.xml
```diff
@@ -0,0 +1,8 @@
+<OverlayData>
+<MarkerCategory  Name="IsFalse">
```
AsherGlick, line 2: Aren't names supposed to be all lower case? I don't actually know if that is a rule or not but I think I see that mostly. idc really.

xml_converter/tests/test_bool.cpp
```diff
@@ -0,0 +1,70 @@
+#include "../src/attribute/bool.hpp"
+#include "../src/packaging_xml.hpp"
```
AsherGlick, line 2: seems strange that the "bool" module that has nothing to do with xml needs to include xml. You should refactor this out.

AsherGlick (changes requested):

xml_converter/test_cases/can_fade_tests/can_fade_is_false.xml
klingbolt: There is no rule they need to be. TacO's documentation states that type and therefore name, "Must not contain any spaces or special characters"

Example code on their page has a name that has capital letters. 

```<OverlayData>

  <MarkerCategory name="Collectible">

    <MarkerCategory name="JournalEntry" iconFile="Data\raid1_achievements.png" behavior="3"/>

  </MarkerCategory>  



</OverlayData>```

File: xml_converter/tests/test_bool.cpp (hunk 2 above)
Comment (klingbolt) [reply xml_converter/tests/test_bool.cpp:2]: Good point. While the test files are for bools, the module is really about comparing xml->xml. Renamed the tests to reflect this.

File: xml_converter/tests/test_bool.cpp [hunk 3]
```diff
@@ -0,0 +1,60 @@
...
+    std::string xml_input = "./test_cases/can_fade_tests/can_fade_is_false.xml";
+    std::string xml_output = "./test_cases/test_output/can_fade_tests/can_fade_is_false.xml";
+    EXPECT_TRUE(compare_files(xml_input, xml_output));
+}
+
+TEST_F(BoolTest, ValueIsZero) {
+    std::string xml_input = "./test_cases/can_fade_tests/can_fade_is_zero.xml";
+    std::string xml_output = "./test_cases/test_output/can_fade_tests/can_fade_is_zero.xml";
+    EXPECT_FALSE(compare_files(xml_input, xml_output));
+}
```
AsherGlick, line 60: These test cases are now not even referencing the code in `src/attribute/bool.*pp`. Should these even be within the cpp unit tests?

Additionally these tests dont test if the code is functioning properly, they test if the files are the same. Meaning that this test binary is now non-functional without the additional bash wrapper script above.

xml_converter/attribute_test_script.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+ATTRIBUTE_TEST_CASES_FOLDER_PATHS=("./test_cases/can_fade_tests")
```
AsherGlick, line 3: Lets get this data dynamically to avoid the toil of having to edit this script over and over every time a new testcase is added.

xml_converter/attribute_test_script.sh
```diff
@@ -0,0 +1,30 @@
...
+        mkdir "$TEST_OUTPUT_DIRECTORY"
+    fi
+
+    while IFS= read -r -d $'\0' file; do
+        ATTRIBUTE_TEST_CASES+=("$file")
+    done < <(find "$ATTRIBUTE_TEST_CASES_FOLDER_PATH" -type f -print0)
+     
+    for test_case in "${ATTRIBUTE_TEST_CASES[@]}"; do
+        file_name=$(basename "$test_case")
+        ./build/xml_converter --input-taco-path "$test_case" --output-taco-path "$TEST_OUTPUT_DIRECTORY/$file_name" 1> /dev/null
```
AsherGlick, line 26: Don't throw away the stdout, it should be compared against what the desired stdout for the test is.

xml_converter/attribute_test_script.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+ATTRIBUTE_TEST_CASES_FOLDER_PATHS=("./test_cases/can_fade_tests")
+
+ATTRIBUTE_TEST_CASES=()
```
AsherGlick, line 5: You dont do anything with this variable in the global scope, and you never reset it within the primary loop. So each iteration of the primary loop you will end up reiterating over this variable which contains all the previous values of previous iterations. If you refactor to remove this variable completely things will be cleaner.

xml_converter/attribute_test_script.sh
```diff
@@ -0,0 +1,30 @@
...
+
+for ATTRIBUTE_TEST_CASES_FOLDER_PATH in "${ATTRIBUTE_TEST_CASES_FOLDER_PATHS[@]}"; do
```
AsherGlick, line 7: Not a hard rule, but using SCREAMING_SNAKE_CASE is sometimes reserved for constants or globals or "something other then a normal variable". in this case `ATTRIBUTE_TEST_CASES_FOLDER_PATH` is a normal variable.

xml_converter/attribute_test_script.sh
```diff
@@ -0,0 +1,30 @@
...
+    if [ ! -d "$ATTRIBUTE_TEST_CASES_FOLDER_PATH" ]; then
+        echo "Folder '$ATTRIBUTE_TEST_CASES_FOLDER_PATH' does not exist."
+        exit 1
+    fi
+
+    attribute_name=$(basename "$ATTRIBUTE_TEST_CASES_FOLDER_PATH")
+    TEST_OUTPUT_DIRECTORY="./test_cases/test_output/$attribute_name"
```
AsherGlick, line 14: The static part of this could be broken out at the top level. The variable name does not need to be capitalized under the same explanation as before.

xml_converter/attribute_test_script.sh
```diff
@@ -0,0 +1,30 @@
...
+        ATTRIBUTE_TEST_CASES+=("$file")
+    done < <(find "$ATTRIBUTE_TEST_CASES_FOLDER_PATH" -type f -print0)
+     
+    for test_case in "${ATTRIBUTE_TEST_CASES[@]}"; do
+        file_name=$(basename "$test_case")
+        ./build/xml_converter --input-taco-path "$test_case" --output-taco-path "$TEST_OUTPUT_DIRECTORY/$file_name" 1> /dev/null
+    done
+done
+
+./build/xml_converter_tests
```
AsherGlick, line 30: Generally a good idea to have a way to run all the tests, but in general software hygiene there should be a way to "run only the integration tests" and "run only the unit tests" and also "run everything". This is because sometimes some tests are much more expensive then other tests (even within one of those two groups) and we want to be able to skip them sometimes.

AsherGlick (changes requested):
Shaping up. Still needs some work though.
You jumped right into some complexity with this shell script, more then I was thinking. Bash is a terrible language as you increase in complexity, so feel free to just use python or something better if you want. Making a shell script was not a requirement.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+class Attribute_Tests:
+    attribute_name: str
+    attribute_path: str
+    tests: List[str]
+    output_result: Dict[str, List[str]]
+    expect_output_to_be_equal: Dict[str, bool]
+    expect_files_to_be_equal: Dict[str, bool]
+
+
+def run_xml_converter(*args):
```
AsherGlick, line 21: this immediately gets converted to a list, just use a list

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+    try:
+        # Build the command to execute the C++ program with the desired function and arguments
+        cmd = [xml_converter] + list(args)
```
AsherGlick, line 24: Lets have the args be broken out in this instead of asking them to be passed in;
eg
```
cmd = [xml_converter, "--input-filez", input_file, "--output-documents", output_file]
```

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # Run the C++ program and capture its output
+        result = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
+
+        # Check if the C++ program executed successfully
+        if result.returncode == 0:
+            return result.stdout
```
AsherGlick, line 31: What if we also have stderr but still have a 0 return code. EG: For warnings.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+        else:
+            return result.stderr
```
AsherGlick, line 33: What if we have stdout with an nonzero return code.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+    except Exception as e:
```
AsherGlick, line 34: If you know exactly what exceptions you are catching this is fine, but catching all exceptions can bite you.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+def are_files_equal(file_path1, file_path2) -> bool:
+    try:
+        with open(file_path1, 'rb') as file1, open(file_path2, 'rb') as file2:
+            content1 = file1.read()
+            content2 = file2.read()
+
+        return content1 == content2
+
+    except (FileNotFoundError, PermissionError):
+        print(f"Error opening file {file_path1} or {file_path2}")
```
AsherGlick, line 48: you can to `as e` at the end of the except and then print the real error too.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+        if result.returncode == 0:
+            return result.stdout
+        else:
+            return result.stderr
+    except Exception as e:
+        print("Error:", e)
+        return None
+
+
+def are_files_equal(file_path1, file_path2) -> bool:
```
AsherGlick, line 39: Lets add  types to these arguments

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+            content2 = file2.read()
+
+        return content1 == content2
+
+    except (FileNotFoundError, PermissionError):
+        print(f"Error opening file {file_path1} or {file_path2}")
+        return False
+
+
+def are_arrays_equal(array1, array2) -> bool:
```
AsherGlick, line 52: Lets add types to these arguments

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+    if len(array1) != len(array2):
+        return False
+
+    for i in range(len(array1)):
+        if array1[i] != array2[i]:
+            return False
+
+    return True
```
AsherGlick, line 60: I am pretty sure if these are arrays of strings then python will do value comparison and not reference comparison so you can just do:
```python
>>> a = [1,2,3]
>>> b = [1,2,3]
>>> a==b
True
```

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+arg_input_xml: str = "--input-taco-path"
+arg_output_xml: str = "--output-taco-path"
+
+chrono_patterns = [
```
AsherGlick, line 66: These are patterns of chronos?

They look like patterns of lines to ignore.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+    "^The taco parse function took [0-9]+ milliseconds to run$",
+    "^The xml write function took [0-9]+ milliseconds to run$",
+    "^The protobuf write function took [0-9]+ milliseconds to run$"
+]
+
+
+if __name__ == "__main__":
```
AsherGlick, line 73: colloquially this block should be at the end of the file and be:
```python
if __name__ == "__main__":
    main()
```

then all this other code is in a new function
```python
def main():
```

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+        expect_output_to_be_equal={
+            "_is_correct": True,
+            "_is_different_file": True,
+            "_is_wrong_output": False,
+        },
+        expect_files_to_be_equal={
+            "_is_correct": True,
+            "_is_different_file": False,
+            "_is_wrong_output": True,
+        }
```
AsherGlick, line 93: can this be something like
```python
attribute: Attribute_Tests = Attribute_Tests(
...
    tests=[
       {output_results="", expect_output_to_be_equal="", expect_files_to_be_equal=""},
       ...
    ]
...
)
```
maybe even allow them to be more then just true/false values
* `expected_output_file_reference: Optional[str]`
* `expected_stdout: Optional[str]`
* `expected_stderr: Optional[str]`

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # Ensure that the test output directory is created
+    if not os.path.exists(attribute.attribute_path + "-output/"):
+        try:
+            os.makedirs(attribute.attribute_path + "-output/")
+        except OSError as e:
+            print(f"Error: {e}")
+
+    for test in attribute.tests:
+        arg1 = attribute.attribute_path + "/" + attribute.attribute_name + test + ".xml"
```
AsherGlick, line 104: give this a better name then `arg1` it looks like an input file argument

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+            "_is_correct": True,
+            "_is_different_file": False,
+            "_is_wrong_output": True,
+        }
+    )
+
+    # Ensure that the test output directory is created
+    if not os.path.exists(attribute.attribute_path + "-output/"):
+        try:
+            os.makedirs(attribute.attribute_path + "-output/")
```
AsherGlick, line 99: break out the variable `attribute.attribute_path + "-output/"` as it is used several times in this code. Having to modify it in multiple places or risk it desyncing is annoying.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+        except OSError as e:
+            print(f"Error: {e}")
+
+    for test in attribute.tests:
+        arg1 = attribute.attribute_path + "/" + attribute.attribute_name + test + ".xml"
+        arg2 = attribute.attribute_path + "-output/" + attribute.attribute_name + test + ".xml"
```
AsherGlick, line 105: give this a better name than `arg2` it looks like an output file argument.

AsherGlick, line 105: use `os.path.join` here instead of string concatenation with `+`.

xml_converter/tests/attribute_testing.py
AsherGlick, line 104: use `os.path.join` here instead of string concatenation with `+`.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+
+        result = run_xml_converter(arg_input_xml, arg1, arg_output_xml, arg2)
+        if result is not None:
+            # Remove lines about how long it took to run
```
AsherGlick, line 109: maybe use "remove noisy lines" or "remove lines we don't care about". This way the comment lives past just caring about lines that have timing information

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+                match_found = False
+                for pattern in chrono_patterns:
+                    if re.search(pattern, line) or line == "":
+                        match_found = True
+                        break
+                if not match_found:
+                    filtered_array.append(line)
+
+            output_test: bool = True
+            file_test: bool = True
```
AsherGlick, line 122: This can just be one variable of `all tests passed` or something like that. And if any test fails then it can set it to false.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+
+        result = run_xml_converter(arg_input_xml, arg1, arg_output_xml, arg2)
+        if result is not None:
+            # Remove lines about how long it took to run
+            split_result = result.split("\n")
+            filtered_array = []
+            for line in split_result:
+                match_found = False
+                for pattern in chrono_patterns:
+                    if re.search(pattern, line) or line == "":
```
AsherGlick, line 115: using .search here means that we will filter out any line that matches any of the regexes to match. This might be fine but also might get us caught if we mess up a regex by missing `^` or `$`.

klingbolt, line 115: Change this to re.fullmatch. This will check the entire string. Just in case something gets added to the end

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,140 @@
...
+                print(f"Expected {attribute.output_result[test]} and got {filtered_array}")
+                output_test = False
+
+            if are_files_equal(arg1, arg2) != attribute.expect_files_to_be_equal[test]:
+                print(f"Files were incorrect for test {attribute.attribute_name}{test}")
+                print(f"Expected {attribute.expect_files_to_be_equal[test]}")
+                print(f"arg1: {arg1}")
+                print(f"arg2: {arg2}")
+                file_test = False
+
```
AsherGlick, line 135: Not necessarily required for either of these things but in general if there is an error because things are different then we will want to know what part of them is different. We can ignore this for now because it is not critical but it will be a useful thing to have later.

AsherGlick (changes requested):
I have not run these tests yet and will probably need to do that in the next round of reviews.

xml_converter/tests/attribute_testing.py
klingbolt: It's possible that we will want to run this command without all of the possible arguments. There are currently 5 that can be used in xml_converter. In this case, I am reading an xml and outputing xml but I plan on adding in reading waypoint files.

AsherGlick: then make the arguments something like:
```python
def run_xml_converter(input_xml: Optional[str], output_xml: Optional[str], input_proto: Optional[str]... etc
```
to abstract the actual functionality for calling the binary. This way you can even re-use the same function or function signature for the library-based tests in the future if that is something that we want to do.

AsherGlick:
I just noticed that there are some merge commits in this branch to that will probably need to be taken care of before merging it in to the main repo.

xml_converter/tests/attribute_testing.py
klingbolt: Makes sense there is a short cut for that

xml_converter/tests/attribute_testing.py
klingbolt: Done

xml_converter/tests/attribute_testing.py
klingbolt: Changed to ```input_xml_path```

xml_converter/tests/attribute_testing.py
klingbolt: Changed to ```output_xml_file```

xml_converter/tests/attribute_testing.py
klingbolt: I think its saver to switch to full match. We aren't going to be filtering too many lines.

xml_converter/tests/attribute_testing.py
klingbolt: I like "remove noisy lines". I'm also going to break it into a separate function since both stdout and stderr might use it.

xml_converter/tests/attribute_testing.py
klingbolt: I like that plan. If any test is false it already prints an error so we don't need to know which test failed later.

xml_converter/tests/attribute_testing.py
klingbolt: I found a module that can generate diffs between files so I have switched to using that.

xml_converter/tests/attribute_testing.py
klingbolt: Done

xml_converter/tests/attribute_testing.py
klingbolt: I realized that since we are going to be explicitly stating what we expect the outputs to be, that things like ```expect_output_to_be_equal = false``` are not testing the attribute themselves but the test program. I don't think that is necessary for the goal of this script. 

I changed the data format to be as follows. 

```
tests": {
            "_is_correct": {
                "output_stdout": [
                    "Loading taco pack ../test_cases/can_fade_tests/can_fade_is_correct.xml"
                ],
                "output_stderr": [],
                "expected_xml_diff": []
            },
```

xml_converter/tests/attribute_testing.py
klingbolt: That makes sense. Making this change

xml_converter/tests/attribute_testing.py
klingbolt: True in isolation the chrono part makes no sense. Changing to ```patterns_for_noisy_lines```

xml_converter/tests/attribute_testing.py
klingbolt: It can. That makes the function unnecessary so it is removed.

xml_converter/tests/attribute_testing.py
klingbolt: done

xml_converter/tests/attribute_testing.py
klingbolt: Good idea. Done

xml_converter/tests/attribute_testing.py
klingbolt: Done

xml_converter/tests/attribute_testing.py
klingbolt: Good point. Since I don't have a specific one in mind that could happen, I'm going to remove this for now. We actually expect some errors if we test for errors.

xml_converter/tests/attribute_testing.py
klingbolt: I'm changing this to return the whole result since we want to have all the information

xml_converter/tests/attribute_testing.py
klingbolt: I'm changing this to return the whole result since we want to have all the information

xml_converter/tests/attribute_testing.py
klingbolt: Done. Changed to reflect the 5 arguments that xml_converter could take in 

```
def run_xml_converter(
    input_xml: Optional[List[str]]= None, 
    output_xml: Optional[List[str]]= None,
    input_proto: Optional[List[str]]= None,
    output_proto: Optional[List[str]]= None,
    split_output_proto: Optional[str]= None,
    ) 
```

klingbolt:
Rebasing to PR #184

PR #180, opened 2023-10-10
XML Files as acceptable input

klingbolt:
The code had an issue where if a xml file was given as an input, the program would segfault. These changes allow an xml file to be passed in as well as allow the user to specify the name of the output file.

### DIFF
--- a/xml_converter/doc/trigger/guid.md
+++ b/xml_converter/doc/trigger/guid.md
@@ -5,6 +5,7 @@ class: UniqueId
 xml_fields: ["GUID"]
 applies_to: ["Icon", "Trail"]
 protobuf_field: guid
+protobuf_type: String
 ---
 A globally unique identifier value to make sure this maker's trigger reset data is always assocaited with this marker and never lost or confused with other markers.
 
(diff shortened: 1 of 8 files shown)


xml_converter/src/xml_converter.cpp
```diff
@@ -89,16 +89,28 @@ void move_supplementary_files(string input_directory, string output_directory) {
     }
 }
 
-void read_taco_directory(string directory, map<string, Category>* marker_categories, vector<Parseable*>* parsed_pois) {
-    vector<string> xml_files = get_files_by_suffix(directory, ".xml");
-    for (const string& path : xml_files) {
-        parse_xml_file(path, marker_categories, parsed_pois);
+void read_taco_directory(string input_path, map<string, Category>* marker_categories, vector<Parseable*>* parsed_pois) {
+    if (has_suffix(input_path, ".xml")) {
```
AsherGlick, line 93: what if I have a folder named "output.xml"

AsherGlick (changes requested):

AsherGlick:
I am also noticing there area bunch of these zero-diff merge commits. Git has a merge option called "fast forward" which allows you to merge a branch that only has new commits onto your current branch without making an additional merge commit.
Screenshot from 2023-10-10 13-29-46

There is no reason to save "how often you synced your local code with the remote code" in git. So if there is not a divergent history that needs to be resolved you should not be adding merge commits within your PRs.

When there is divergent history (like the painful and inevitable merge we will need to do for master->xml_converter) that will have a merge commit within the pr.

PR #182, opened 2023-10-15
Json for test data

klingbolt:
Dependent on #178 

This takes the data that was stored in the dataclass and stores it in a JSON file. We will be able to easily add to the JSON new tests and attributes.


klingbolt:
Changes covered in PR https://github.com/AsherGlick/Burrito/pull/184

PR #183, opened 2023-10-18
Allows files and directories to inputs and handles them based on type

klingbolt:
The code had an issue where if a xml file was given as an input, the program would segfault. These changes allow an xml file to be passed in as well as allow the user to specify the name of the output file.


xml_converter/src/xml_converter.cpp
```diff
@@ -89,16 +90,40 @@ void move_supplementary_files(string input_directory, string output_directory) {
...
+        cout << "Error: Unknown file type" << endl;
     }
 }
 
-void write_taco_directory(string directory, map<string, Category>* marker_categories, vector<Parseable*>* parsed_pois) {
+void write_taco_directory(string output_path, map<string, Category>* marker_categories, vector<Parseable*>* parsed_pois) {
     // TODO: Exportion of XML Marker Packs File Structure #111
-    string xml_filepath = directory + "xml_file.xml";
+    string xml_filepath;
+    if (filesystem::is_regular_file(output_path)) {
```
AsherGlick, line 116: What if the output file does not exist how do I differentiate between wanting to create a file or a directory?

Case 1:
* "I want to write to the file `output.xml`"

Case 2:
* "I want to write to the folder `output.xml` (which would then get a file saved to `output.xml/xml_file.xml`")

xml_converter/src/xml_converter.cpp
```diff
@@ -89,16 +90,40 @@ void move_supplementary_files(string input_directory, string output_directory) {
...
+        if (has_suffix(output_path, ".xml")) {
+            xml_filepath = output_path;
+        }
```
AsherGlick, line 119: What if I want to write to my backup xml file `my_xml_data.xml.backup` ?

AsherGlick (changes requested):

xml_converter/src/xml_converter.cpp
klingbolt: Right now, the program does not make any directories that do not exist. Prior to this PR, It just assumes that any path is a valid directory. This might be better addressed in a separate PR.  

I think I have a workable solution though and I'd like to get your feedback before I submit it. 

1. Default to writing a folder
- The main use of this program is to create an exportable folder so it should probably default to something that is easy to zip and share.
2. Separate argument for file or dir outputs
- We want to allow an option of exporting to a file instead so we Introduce a new argument separate from ````output-taco-path``` for writing to an XML file rather than a dir (i.e. ```output-xml-path```
3a. Make the directory if using ```output-taco-path```
- Either in this function or the one before, add a check to see if the directory exists and try to make it if it does not.
3b. Have a new function for ```write_taco_output()``` 
- This would separate out the logic for files and dirs and make it clear which output is being requested

xml_converter/src/xml_converter.cpp
klingbolt: Good point. I'm changing it to not require the xml at the end

xml_converter/src/xml_converter.cpp
AsherGlick: break out the pros and cons of each strategy. Include user stories for each use case and variations for how those would work alongside each strategy. Don't forget to include rational for why a user story would exist, "Fred wants to export an xml file using korean ligatures because :shrug: " should be explored to see if Fred even really wants to do that.

klingbolt: I think I've boiled this down to two options. Either we can make the arguments simple and use educated guesses as to the file type wanted or we can adjust arguments to allow an explicit behavior. After weighing the options, I think that option 1 is the smoothest user experience and that the use cases where the default behavior is not wanted are rare enough that the extra complexity isn't worth the pros.

1. 

Make educated guesses and have a default behavior:

    Only two clues they can communicate to us. `has_suffix("/")` or `has_suffix(".xml")

    If it ends with /, it's likely a folder.

    If it ends with .xml, it's likely an XML file.

    If neither, default behavior is to try and write a folder.

Pros:

    Simple context clues and can give clear expected outcomes. Could add a flag to change default behavior later.

Cons:

    Would need to document the default behavior.

Case 0:
"I want to write a file or dir that exists."

    This works as expected.

Case 1:
"I want to write to the file output.xml."

    This would output an XML file.

Case 2:
"I want to write to the folder output.xml (which would then get a file saved to output.xml/xml_file.xml)."

    This would also output an XML file. If they want to do this, they need to first make a directory called output.xml.

Case 3:
"I want to write to the file output.xml.backup"

    This would output a directory with and XML file. If they want to do this, they need to first make a file called output.xml.backup

2. 

Change arguments to specify:

    Need to specify which output format to use.

    `--output-taco-path` makes a folder.

    `--output-xml-path` makes a file.

Pros:

    Most explicit method for stating the output.

Cons:

    Requires extra knowledge of arguments, could lead to user errors, ties "taco-path" to a directory.

Case 0:
"I want to write a file or dir that exists."

    Need to determine if it is a file or directory and choose the correct argument. Could lead to user error.

Case 1:
"I want to write to the file output.xml."

    Need to specify --output-xml-path.

Case 2:
"I want to write to the folder output.xml (which would then get a file saved to output.xml/xml_file.xml)."

    Need to specify --output-taco-path.

AsherGlick: For the initial argument:
* What source are you using to determine that the non-default behavior is rare


## 1. Make educated guesses and have a default behavior:


For the Pros
* isn't requiring the user to know context clues to use the program a con?

For the cons
* in addition to making documentation you also need the users to read the documentation
* an additional con is that users must create the output file manually beforehand if they want a folder ending in .xml or a file ending in anything but .xml
* an additional con is that the same command run at different points in time may behave differently based on the state of the output filesystem. Even if all the inputs and command are the same.


## 2. Change arguments to specify:


For the cons
* This is claimed to lead to user errors, is this more or less likely to cause user errors then the other option

For Case 0
* This same user error capable decision needs to be made for the other option as well for the user to determine if they should create/delete a file/directory before running this command.


## Additional points to consider:
* How will this interaction change when we eventually write the `.taco` files.
* How will this interaction change when we eventually allow splitting based on category
* For the situations where there may be user errors, what are the methods of discovery/debugging the users take to correct the error

xml_converter/src/xml_converter.cpp
```diff
@@ -89,16 +90,47 @@ void move_supplementary_files(string input_directory, string output_directory) {
...
+    else {
+        if (filesystem::create_directory(output_path)) {
+            if (!has_suffix(output_path, "/"))
+                output_path += "/";
+            xml_filepath = output_path + "xml_file.xml";
+        }
+        else {
+            cout << "Error: " << output_path << "is not a valid directory path" << endl;
+        }
+    }
```
AsherGlick, line 133: This logic can be compressed to:
```
if the directory does not exist:
    create it()
write to this directory()
````
with whatever error checking is necessary interspersed

The current pattern has a bunch of duplicated logic that can be deduplicated.

**Review by AsherGlick — CHANGES_REQUESTED** [review]

File: xml_converter/src/xml_converter.cpp [hunk 4]
```diff
@@ -89,16 +90,37 @@ void move_supplementary_files(string input_directory, string output_directory) {
...
+void write_taco_directory(string output_path, map<string, Category>* marker_categories, vector<Parseable*>* parsed_pois) {
     // TODO: Exportion of XML Marker Packs File Structure #111
-    string xml_filepath = directory + "/xml_file.xml";
+    string xml_filepath;
+    if (!has_suffix(output_path, "/")) {
+        output_path += "/";
+        xml_filepath = output_path + "xml_file.xml";
+    }
+    else {
+        xml_filepath = output_path;
```
AsherGlick, line 116: I dont understand why this does not have xml_file.xml

AsherGlick (approved):

PR #184, opened 2023-10-22
Attribute test harness

klingbolt:
This python script and tests can be used to validate the output of ```xmlconverter.cpp``` against a known value so that we can catch any issues. 

This PR only takes in an xml and writes to xml. The plan is to establish the framework here so that more tests can be written.

This PR is a rebase of PR #178 and makes it obsolete.


xml_converter/test_cases/test_expected_outputs.json
```diff
@@ -0,0 +1,38 @@
+{
+    "can_fade": {
+        "path": "../test_cases/can_fade_tests",
+        "tests": {
```
AsherGlick, line 4: From experience in situations like this, using a dict with a list of items in the format `{myname: {mydata}}` always ends up being more pain then `[{name: myname, ...mydata...}]` in the long term. You dont need to be able to lookup these tests by name so having it as a key it not beneficial and you do need to iterate over the data and a dict is unsorted. (Technically python does guarentee ordering of dicts be the same as the input order but this is not true in all languages (like cpp which will sort them alphabetically, or go which goes out of its way to sort them randomly)). You might also want to do this for the top level testgroups too.

xml_converter/test_cases/test_expected_outputs.json
```diff
@@ -0,0 +1,38 @@
...
+            "_is_correct": {
```
AsherGlick, line 5: Why are these names underlined?

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,165 @@
...
+json_file_path: str = "../test_cases/test_expected_outputs.json"
+
+
+@dataclass
+class XMLConverter_arguments:
+    arg_input_xml: str = "--input-taco-path"
+    arg_output_xml: str = "--output-taco-path"
+    arg_input_proto: str = "--input-waypoint-path"
+    arg_output_proto: str = "--output-waypoint-path"
+    arg_split_proto: str = "--output-split-waypoint-path"
```
AsherGlick, line 21: This is being used as a static dictionary for a magic string. A better convention here is to just make these top level constant variables
```
arg_input_xml: Final[str] = "--input-taco-path"
...
```
no need to make a dataclass for static data. These are also only used once and not a value that a programmer would really need to care about changing too so you might not need them at all.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,165 @@
...
+import argparse
+import difflib
+import json
+import subprocess
+import re
+import os
+from typing import List, Optional
+
+# Path to compiled C++ executable
+xml_converter: str = "../build/xml_converter"
```
AsherGlick, line 11:
```suggestion
xml_converter_binary_path: str = "../build/xml_converter"
```

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,165 @@
...
+        differ = difflib.Differ()
+        diff = list(differ.compare(content1, content2))
+        filtered_diff: List[str] = []
+        for line in diff:
+            if line.startswith("+ ") or line.startswith("- "):
+                filtered_diff.append(line)
+
+        return filtered_diff
+
+    except (FileNotFoundError, PermissionError) as e:
```
AsherGlick, line 67: PermissionError should probably still throw a error.
FileNotFoundError should probably return a complete file diff of the other, depending on which file was not found. And maybe throw an error still if both are not found.
If you even need to catch errors here at all.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,165 @@
...
+patterns_for_noisy_lines = [
+    "^The taco parse function took [0-9]+ milliseconds to run$",
+    "^The xml write function took [0-9]+ milliseconds to run$",
+    "^The protobuf read function took [0-9]+ milliseconds to run$",
+    "^The protobuf write function took [0-9]+ milliseconds to run$",
+    "^$"
+]
+
+
+def remove_noisy_lines(array: List[str]) -> List[str]:
```
AsherGlick, line 82:
```suggestion
def remove_noisy_lines(lines: List[str]) -> List[str]:
```
this function is named to do a specific thing, make the args as specific

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,165 @@
...
+    filtered_array = []
+    for line in array:
```
AsherGlick, line 84:
```suggestion
    for line in lines:
```

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,165 @@
...
+        print(f"JSON decoding error: {e}")
+    except Exception as e:
+        print(f"An error occurred: {e}")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="An example script with an optional argument.")
+    parser.add_argument("-p", "--print", help="Prints the results rather than comparing them to a file", action="store_true")
+
+    args = parser.parse_args()
```
AsherGlick, line 164: This can just go inside main() if you want to avoid passing around an argparse object.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,165 @@
...
+            if re.fullmatch(pattern, line):
+                match_found = True
+                break
+        if not match_found:
+            filtered_array.append(line)
+    return filtered_array
+
+
+def main(args: argparse.Namespace) -> None:
+    try:
```
AsherGlick, line 96: Don't wrap the entire python script in a `try`, even if you are only handling catching certain errors. Using a try statement for anything more then the bare minimum is begging for invisible trouble.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,165 @@
...
+                    all_tests_passed = False
+
+                if all_tests_passed:
+                    print(f"Success: test {attribute_name}{test}")
+
+    except FileNotFoundError:
+        print(f"The file '{json_file_path}' does not exist.")
+    except json.JSONDecodeError as e:
+        print(f"JSON decoding error: {e}")
+    except Exception as e:
```
AsherGlick, line 156: This single line catches every possible error that could possibly occur anywhere inside this entire executable except for `FileNotFoundError`s, `JSONDecodeError`s, and the three lines of code that are outside of the `main()` function. Lets not do that.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,165 @@
...
+        cmd += [XMLConverter_arguments.arg_input_proto] + input_proto
+    if output_proto:
+        cmd += [XMLConverter_arguments.arg_output_proto] + output_proto
+    if split_output_proto:
+        cmd += [XMLConverter_arguments.arg_split_proto] + [split_output_proto]
+
+    # Run the C++ program and capture its output
+    result = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
+
+    return result
```
AsherGlick, line 49: Wrapper functions like this I tend to return custom data instead of library return values. These functions are wrappers in the first place so returning a custom return type means that we can control both the input and the output function signatures and do any of the grunt work needed all within the wrapper function.
In this case I would return
* stdout
* stderr
* return code

Because those are the things that our particular program might care about outside of this function. You can return these as a tuple, or if you want to be extra, as a dataclass. Both options have pros and cons. Tuples can be implicitly deconstructed, dataclasses can handle more complex types but require a forward declaration and will only be used in a single place as one return value. If I were writing this I would return a tuple.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,165 @@
...
+
+        for attribute_name in data.keys():
+            print(attribute_name)
+            attribute_data = data[attribute_name]
+
+            # Ensure that the test output directory is created
+            output_dir_path = attribute_data["path"] + "-output/"
+            if not os.path.exists(output_dir_path):
+                try:
+                    os.makedirs(output_dir_path)
```
AsherGlick, line 108: `os.makedirs` has an argument `exist_ok` which allows you to ignore if the directory exists or not.

xml_converter/test_cases/test_expected_outputs.json
```diff
@@ -0,0 +1,38 @@
+{
+    "can_fade": {
+        "path": "../test_cases/can_fade_tests",
```
AsherGlick, line 3: this json file is in the `test_cases` directory. A relative path from where this file is located would be `can_fade_tests` because the `../test_cases` goes up a folder and then back in.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,165 @@
...
+        print(f"The file '{json_file_path}' does not exist.")
+    except json.JSONDecodeError as e:
+        print(f"JSON decoding error: {e}")
+    except Exception as e:
+        print(f"An error occurred: {e}")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="An example script with an optional argument.")
+    parser.add_argument("-p", "--print", help="Prints the results rather than comparing them to a file", action="store_true")
```
AsherGlick, line 162: This is not quite the same, but when printing out additional data the colloquial flag is `-v` for `--verbose`. Often there will be multiple levels of verbosity so sometimes you will see command line flags that read `-vvvvvv` for "just make it super dee duper verbose please"

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,165 @@
...
+                stderr = remove_noisy_lines(result.stderr.split("\n"))
+                xml_diff = compare_files(input_xml_path, output_xml_path)
+
+                # Prints the results rather than comparing them to a file
+                if args.print:
+                    print(f"Test {attribute_name}{test}")
+                    print(f"'output_stdout' : {json.dumps(stdout)}")
+                    print(f"'output_stderr' = {json.dumps(stderr)}")
+                    print(f"'expected_xml_diff' = {json.dumps(xml_diff)}")
+                    continue
```
AsherGlick, line 129: Why is printing the output *instead* of comparing the output useful?

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,165 @@
...
+
+                all_tests_passed: bool = True
+                expected_results = attribute_data["tests"][test]
+
+                if stdout != expected_results["output_stdout"]:
+                    print(f"Output did not match for test {attribute_name}{test}")
+                    print(f"Expected stdout {expected_results['output_stdout']} and got {stdout}")
```
AsherGlick, line 136: If a test fails lets print out the diff of the expected and the actual. Same for the others.

AsherGlick (changes requested):

AsherGlick:
Go ahead and close the PRs that are being rebased by this one.
Does this also cover #182 ?

xml_converter/tests/attribute_testing.py
klingbolt: I was unsure what the right word would be here. Changed to ```-v``` and ```--verbose```

xml_converter/tests/attribute_testing.py
AsherGlick: They could be. But Final is not important enough for me to prevent a merge over.

xml_converter/tests/test_expected_outputs.json
```diff
@@ -0,0 +1,36 @@
...
+        "tests": [
+            {
+                "name": "is_correct", 
+                "expected_stdout" : [
+                    "Loading taco pack ../test_cases/can_fade/inputs/can_fade_is_correct.xml"
+                ],
+
+                "expected_stderr": [],
+                "expected_returncode" : 0
+            },
```
AsherGlick, line 13: lets add a field here for a path to the expected output file.

xml_converter/tests/test_expected_outputs.json
```diff
@@ -0,0 +1,36 @@
...
+                    "Loading taco pack ../test_cases/can_fade/inputs/can_fade_is_supported_value_not_default.xml"
+                ], 
+                "expected_stderr": [],
+                "expected_returncode" : 0
+            },
+            {
+                "name": "unsupported_value",
+                "expected_stdout": [
+                    "Loading taco pack ../test_cases/can_fade/inputs/can_fade_unsupported_value.xml",
+                    "\u001b[31;1mError: \u001b[0m\u001b[1mFound a boolean value that was not a '1', '0', 'true', or 'false'\u001b[0m",
```
AsherGlick, line 26: Lets strip out the color escape codes from these strings in the test harness so we dont need to include them in the expected stdout.

xml_converter/tests/attribute_testing.py
```diff
@@ -0,0 +1,164 @@
...
+            result = run_xml_converter(input_xml=[input_xml_path], output_xml=[output_xml_path])
+
+            # Remove noisy lines
+            stdout: List[str] = remove_noisy_lines(result[0].split("\n"))
+            stderr: List[str] = remove_noisy_lines(result[1].split("\n"))
+            returncode: int = result[2]
+
+            # Prints the results rather than comparing them to a file
+            if args.verbose:
+                print(f"\033[94mTest {attribute_name}_{test['name']}\033[0m")
```
AsherGlick, line 121: Lets remove the color codes being printed out for now to reduce the number of changes. You can add them back in in a different PR. When you do you can set variables equal to them so you can more easily understand what this string says. You could also turn them into functions like `red_text('Test {attribute_name)_{test['name']')` then

AsherGlick (changes requested):

xml_converter/test_cases/can_fade/inputs/can_fade_is_correct.xml
```diff
@@ -1,8 +1,8 @@
 <OverlayData>
-<MarkerCategory  Name="iscorrect">
```
AsherGlick: This should probably be a value we expect to receive. Otherwise when we write the test that checks to make sure all category names exist and all poi types exist we might have to change all the tests. You can use a single name across all tests though such as `mycategory`.

xml_converter/tests/test_expected_outputs.json
```diff
@@ -3,16 +3,17 @@
         "attribute_name": "can_fade",
         "tests": [
             {
-                "name": "is_correct", 
+                "name": "is_correct",
+                "xml_expected_output_file_name": "can_fade_false.xml",
```
AsherGlick, line 7: Thoughts?
```suggestion
                "expected_xml_output_file_name": "can_fade_false.xml",
```

AsherGlick (changes requested):

xml_converter/test_cases/can_fade/inputs/can_fade_is_correct.xml
klingbolt: That's a good idea. Adding

AsherGlick (approved):

PR #185, opened 2023-10-25
Merging master into XML converter

AsherGlick:
Resolved the merge conflicts between xml converter and master so that xml converter can more easily be merged back into master when it is completed.

When merged, this PR should be fastforwarded, no need to create a merge commit for a merge commit.


Spatial.gd
```diff
@@ -255,6 +321,10 @@ func decode_context_packet(spb: StreamPeerBuffer):
 	# this to just be a radian to degree conversion.
 
 	if self.map_id != old_map_id:
+		print("New Map")
+
+		print("Saving Old Map")
+		self.markerdata[str(old_map_id)] = data_from_renderview()
```
klingbolt, line 327: This line will crash. Markerdata is now defined on line 365 as ```var markerdata = Waypoint.Waypoint.new()```

Spatial.tscn
```diff
@@ -600,125 +601,158 @@ disabled = true
 text = "Reverse Path Direction"
 
 [node name="SettingsDialog" type="WindowDialog" parent="Control/Dialogs"]
+visible = true
```
klingbolt, line 604: This should be false. Godot gives a node configuration warning that pop ups will be hidden on run time

klingbolt, line 604:
```suggestion
visible = false
```

Spatial.gd
klingbolt, line 327:
```suggestion
```

klingbolt (changes requested):

Spatial.tscn
AsherGlick: accepting suggestion to remove the line, it will be re-added when saving is recreated.

klingbolt (approved):

AsherGlick:
Woops, it looks like I made a merge commit. Oh well, not going to fix it now we will just live.

Issue #190, opened 2023-10-30
[XML Converter] MultiFlag Classes incorrect output


klingbolt:
The filter classes are currently outputting incorrect information. The filters should have an output with a comma between.

e.g. `Mount="raptor,springer,beetle"` but we are currently outputting `Mount="raptorspringerbeetle"`

AsherGlick:
Fixed with #191

PR #191, opened 2023-10-31
Adds a comma inbetween the flags when writing to xml

klingbolt:
Address issue #190 

The filter classes are currently outputting incorrect information. The filters should have an output with a comma between.

e.g. Mount="raptor,springer,beetle" but we are currently outputting Mount="raptorspringerbeetle"


xml_converter/src/attribute/festival_filter_gen.cpp
```diff
@@ -59,27 +59,34 @@ FestivalFilter parse_festival_filter(rapidxml::xml_attribute<>* input, vector<XM
...
     if (value->none == true) {
-        output = output + "none";
+        flag_values.push_back("none");
+    }
+    string output = "";
+    for (size_t i = 0; i < flag_values.size(); ++i) {
+        output += flag_values[i];
+        if (i < flag_values.size() - 1) {
+            output += ",";
+        }
```
AsherGlick, line 89: use the stringhelper join function instead here. It does exactly what you want and (will soon have) an optimization to preallocate the full size of the output string instead of reallocating each iteration of the loop.

AsherGlick (changes requested):

AsherGlick:
Also add a new testcase for this change to prove the code is doing what you expect.

AsherGlick (approved):

PR #220, opened 2023-11-14
Tests for Fade Near and Far

klingbolt:
Adding integration tests for FadeNear and FadeFar. The two numbers are a pair and should be expected to both be used if one or the other is present.


klingbolt:
These tests are written so that Distance Fade Start and End are only 0 or positive and Fade Near is less than Fade Far. #221 was opened to track adding these requirements to the parser.

AsherGlick (approved):

PR #225, opened 2023-11-14
Dynamic icon positon depending on ui-scale

Masgalor:
This PR aims to change the position of the main button/icon according to the used ui-scale.
Nothing will be scaled or otherwise resized, all this PR does is recalculating the position.

#### What needs to happen to move the button/icon and the clickable are.
 - The left- and the right-margin must be changed. https://github.com/AsherGlick/Burrito/blob/a344bdfcfa738ca764ffe677c5f7ba10217b15db/Spatial.tscn#L74-L75
 - The function `set_minimal_mouse_block()` must be called _once_ after changing the margins. https://github.com/AsherGlick/Burrito/blob/a344bdfcfa738ca764ffe677c5f7ba10217b15db/Spatial.gd#L121

#### How to calculate the margins.
Every button has a fixed width depending on the ui-scale.
If we know the width of one button we can calculate the margin of every n..th button.
So we can preset the button width for every scale, we can also preset the desired position of our button. (e.g. if the native ui has 10 buttons we want to be on position 11.)
After reading the ui-scale from the link, it can be used to select the appropriate base width and calculate the final position.

#### Possible conflicts.
When the function `set_minimal_mouse_block()` is called, the clickable area of all open windows will be deleted.
So the user will see the windows but can no longer interact with them.

### DIFF
--- a/Spatial.gd
+++ b/Spatial.gd
@@ -51,10 +51,10 @@ var is_transient:bool = false
 # The ui-size as read from the link can have the values [0=small; 1=normal; 2=large; 3=larger]
 var ui_size = 1
 # This array holds the width of one item for every ui-scale
-var icon_size_preset = [26.5, 29.5, 33.0, 36.0] # 0=small; 1=normal; 2=large; 3=larger
+const icon_size_preset = [26.5, 29.5, 33.0, 36.0] # 0=small; 1=normal; 2=large; 3=larger
 # The position indicates how many buttons will be there.
 # e.g. if the native ui has 10 buttons we want to be on position 11.
-var button_position = 11 # Place the icon/button on n..th position
+const button_position = 11 # Place the icon/button on n..th position
 
 # Called when the node enters the scene tree for the first time.
 func _ready():
--- a/Spatial.gd
+++ b/Spatial.gd
@@ -48,6 +48,8 @@ var x11_window_id_burrito: int
 var is_transient:bool = false
 
 # Variables that store informations about ui scaling
+# The ui-size as read from the link can have the values [0=small; 1=normal; 2=large; 3=larger]
+var ui_size = 1
 # This array holds the width of one item for every ui-scale
 var icon_size_preset = [26.5, 29.5, 33.0, 36.0] # 0=small; 1=normal; 2=large; 3=larger
 # The position indicates how many buttons will be there.
@@ -330,9 +332,15 @@ func decode_context_packet(spb: StreamPeerBuffer):
 	# The left- and right-margin can be calculated from the ui-scale combined with the preset icon width and the desired position.
 	# set_minimal_mouse_block() should be called only once, if it is called while any burrito windows is open it will become unclickable.
 	# TODO: Check if the calculated position is inside the window.
-	if $Control/GlobalMenuButton.margin_left != icon_size_preset[identity["uisz"]] * button_position:
-		$Control/GlobalMenuButton.margin_left = icon_size_preset[identity["uisz"]] * button_position
-		$Control/GlobalMenuButton.margin_right = (icon_size_preset[identity["uisz"]] * (button_position + 1))
+
+	ui_size = identity["uisz"]
+	# If the value is outside of the expected range use the "normal" size.
+	if (ui_size < 0) or (ui_size > 3):
+		ui_size = 1
+
+	if $Control/GlobalMenuButton.margin_left != icon_size_preset[ui_size] * button_position:
+		$Control/GlobalMenuButton.margin_left = icon_size_preset[ui_size] * button_position
+		$Control/GlobalMenuButton.margin_right = (icon_size_preset[ui_size] * (button_position + 1))
 		set_minimal_mouse_block()
 
 	if self.map_id != old_map_id:
--- a/Spatial.gd
+++ b/Spatial.gd
@@ -51,7 +51,7 @@ var is_transient:bool = false
 # The ui-size as read from the link can have the values [0=small; 1=normal; 2=large; 3=larger]
 var ui_size: int = 1
 # This array holds the width of one item for every ui-scale
-const icon_size_preset = [26.5, 29.5, 33.0, 36.0] # 0=small; 1=normal; 2=large; 3=larger
+const icon_size_preset = {0: 26.5, 1: 29.5, 2: 33.0, 3: 36.0} # 0=small; 1=normal; 2=large; 3=larger
 # The position indicates how many buttons will be there.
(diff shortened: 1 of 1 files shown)


AsherGlick:
Once you have implemented everything you want to with this, are happy with the functionality, and convert this from a draft pr to a regular pr, then we will review this. It might take some time before we get to it, depending on our personal schedules and free cycles, but we are aware of it.

Spatial.gd
```diff
@@ -47,6 +47,15 @@ var taco_parser: TacoParser
 var x11_window_id_burrito: int
 var is_transient:bool = false
 
+# Variables that store informations about ui scaling
+# The ui-size as read from the link can have the values [0=small; 1=normal; 2=large; 3=larger]
+var ui_size = 1
+# This array holds the width of one item for every ui-scale
+const icon_size_preset = [26.5, 29.5, 33.0, 36.0] # 0=small; 1=normal; 2=large; 3=larger
```
klingbolt, line 54: When we were trying to calculate the position before we found it tricky. We found the best position value for normal to be 323. I think we should just store the full position value instead of trying to calculate it. If down the line, GW2 adds another button, we can manually change the values.

Spatial.gd
```diff
@@ -319,6 +328,27 @@ func decode_context_packet(spb: StreamPeerBuffer):
...
 
+	# Calculations to dynamically place the main icon/button
+	# The left- and right-margin can be calculated from the ui-scale combined with the preset icon width and the desired position.
+	# set_minimal_mouse_block() should be called only once, if it is called while any burrito windows is open it will become unclickable.
+	ui_size = identity["uisz"]
+	# If the value is outside of the expected range use the "normal" size.
+	if (ui_size < 0) or (ui_size > 3):
+		ui_size = 1
+
+	var button_position_max = floor(OS.window_size.x / icon_size_preset[ui_size]) - 1
```
klingbolt, line 339: I'm not sure this is the correct solution. Guild Wars 2 seems to shrink the UI in small windows. I have some examples below. 

image
image

klingbolt (changes requested):

Spatial.gd
Masgalor: I am not sure with this one, my values might not be optimal but technically there is no big difference.
If I would change my value for `size: normal` to 29.36 we would end with 322.96 for our button.

If ANET adds another button we need to change one constant, if we do everything absolute we would need to change margin_left and margin_right four times each.
Doing it my way would also allow us to let the user change the position with a setting like, move button one step to the right / one step to the left.

That said, if you think it's not worth I will rewrite it.

Spatial.gd
Masgalor: You are right I did not think about a movable window, my point of view was a borderless fullscreen window.
If the position for our button ends up outside the window (for whatever reason) it would not be visible, with that limit it would be appear in the right corner of the screen regardless.

I will try to find a way to keep the button inside the window even if it is moved around.

AsherGlick: IIRC there was some discussion on the blishud discord at some point about how this sizing algorithm works. I cant remember where but I think it is documented... somewhere.

Spatial.gd
AsherGlick: I'm inclined to agree with @klingbolt that it would be preferred over what is in this pr now. But there is probably an even better solution if you want to put in the legwork.

Something about having fractional values here sets off some alarms form me, at least for 'normal', I could maybe see an argument for why they would exist for small, maybe.

It is possible that the position of the icon is also not as simple as just a width, it could be there is an additional padding value somewhere before/after/between each icon, or a fixed left padding offset for just the first one. By finding the icon assets and lining them up against screenshots we can figure out what the exact coordinates are for each icon and then use those offsets to determine the correct algorithm. We can ignore all of that if we just used a fixed value.

Like you say, the downside to a single fixed value is if anet ever adds another button. Looking at the history of anet adding new buttons to gw2, I think this event will be *very* rare, and editing 4 numbers is not too much more work.

Spatial.gd
Masgalor: Ok, I thought about this and come to the conclusion that running GW2 in a partial window needs a lot of work outside the scope of this PR.
 - The position of burrito does not follow the GW2 window
 - We need to work in an offset to compensate the titlebar
 - Dynamic scaling of elements

Screenshot_20231120_221219

So, we can either keep this check, because it does what I want it to do as long as GW2 is running fullscreen or we remove it because it is not useful.

Spatial.gd
```diff
@@ -319,6 +328,26 @@ func decode_context_packet(spb: StreamPeerBuffer):
...
+	var button_position_max = floor(OS.window_size.x / self.icon_size_preset[self.ui_size]) - 1
+	var button_margin_left = $Control/GlobalMenuButton.margin_left
+	# Make sure the expected position is inside the window.
+	if (self.button_position > button_position_max):
+		button_margin_left = self.icon_size_preset[self.ui_size] * button_position_max
+	else:
+		button_margin_left = self.icon_size_preset[self.ui_size] * self.button_position
+
+	$Control/GlobalMenuButton.margin_left = button_margin_left
+	$Control/GlobalMenuButton.margin_right = button_margin_left + self.icon_size_preset[self.ui_size]
```
AsherGlick, line 347: Setting the right margin might not be necessary, the right margin might be being ignored here due to the anchors. Not really important but if it is not necessary might as well remove it.

Spatial.gd
AsherGlick: If it were my pr I would just remove this check and assume the window is big enough, then revisit it when we have a sticky position

Spatial.gd
Masgalor: Not setting the right margin caused the button to disappear on scale `large` and `largest` .
I will check this again.

Masgalor: The function `set_minimal_mouse_block()` will not work if the left margin is greater than the right margin.
So we have to set both or rework said function.

AsherGlick: Makes sense. Ignore this comment then.

Spatial.gd
Masgalor: Ok, i still have a different opinion but i am happy to implement the requested changes.
That means we want see something like this, right?
```
const button_margin {
	0: {"left": 292, "right": 318},
	1: {"left": 323, "right": 352},
	2: {"left": 363, "right": 396},
	3: {"left": 396, "right": 432}
}

$Control/GlobalMenuButton.margin_left = button_margin[self.ui_size]["left"]
$Control/GlobalMenuButton.margin_right = button_margin[self.ui_size]["right"]
```

AsherGlick (approved):

Issue #151, opened 2023-07-31
[Proto Optimization] Optomizing the proto file


klingbolt:
The current `.proto` file was written based on the way data is stored in the XML files, but we aren't restricted to that. The following is a list of improvements that can be made to either improve data storage in the protobuf formatted files or improve the performance of the Godot code. Many on this list will be changing the code so that a default value is always stored as 0 if possible and that both the XML Converter and Burrito handle the default value correctly. This list includes the current default values if applicable.

- [x] Make Icon and Trail elements of Category #152 
- [x] Reduce display_name and name to one variable #202
- [x] Make texture_path a field of Waypoint to reduce duplicate texture strings #250
- [x] Change RGBAColor from message to fixed32 (Default  = #FFFFFFFF) #312
- [x] GUID change from message to bytes #179
- [x] Defaults are Bool
  - [x] Can Fade (Default  = 1 ) #210
  - [x] Default Toggle (Default  = 1 ) #347
  - [x] Render_ingame (default true) #350
  - [x] Render_on_map/minimap (default true) #351 #352
  - [x] Scale On Map With Zoom (default = 1) #348

The following fields could be improved if the default value is stored in the proto file as 0 where both Burrito and xml_converter know that "0" should be interpreted as the default value. For now, these values are not implemented in burrito, so we are going to comment them out in the proto. They can be re-enabled once a solution is found to keep track of default values in xml_converter. 
- [ ] Min and Max
  - [ ] Size (Default = 5 for min and 2048 for max)
  - [ ] Fade (Defaults = 700 for near and 900 for far)
- [ ] Set values to 0 if equal to default
  - [ ] Scale (Default = 1)
  - [ ] Icon Size (Default = 1)
  - [ ] Animation Speed (change from default= 1 to default = 0)
  - [ ] Height Offset (Default = 1.5)
  - [ ] Map Display Size (Default = 20)
  - [ ] Trigger Range (default = 2)

AsherGlick:
Some brainstorming:
* `Category.name` - Is not used in the proto now that icons and trails are nested
  * `Category.display_name` should be used to infer what `name` should be when writing to xml without name otherwise set
    * `Category.display_name` should also be renamed to `Category.name`

* `Icon.texture_path`, `Trail.texture_path` - is possibly a highly duplicated string, existing on every icon and trail that uses the same texture
    * Replace with uint32 index into a top-level array of texture path objects, useful if there is extra info that should be included alongside the texture
[...]

PR #228, opened 2023-11-15
Anim speed

klingbolt:
The protobin is not added here as it will be changed by #151


xml_converter/intigration_tests/inputs/xml_animation_speed/xml_file.xml
```diff
@@ -0,0 +1,13 @@
...
+    <MarkerCategory DisplayName="My Category" Name="mycategory">
+    </MarkerCategory>
+
+    <POIs>
+        <Trail AnimSpeed="0" Type="mycategory" />
+        <Trail AnimSpeed="3.14" Type="mycategory" />
+        <Trail AnimSpeed="123.456" Type="mycategory" />
+        <Trail AnimationSpeed="0" Type="mycategory" />
+        <Trail AnimationSpeed="3.14" Type="mycategory" />
+        <Trail AnimationSpeed="123.456" Type="mycategory" />
```
AsherGlick, line 11: Can this be negative? With the animation running reverse?

AsherGlick (changes requested):

xml_converter/intigration_tests/inputs/xml_animation_speed/xml_file.xml
klingbolt: I don't believe it can. The BlishHud docs do not specify how they implement it but in Godot, there is a property of the animation player called `playback_speed` which also has a default of 1.0 that is a speed scaling ratio. That is likely the field that would be edited.  This also does not say in the documentation if it can be negative.

AsherGlick: https://docs.godotengine.org/en/3.3/classes/class_animationplayer.html#class-animationplayer-method-play

xml_converter/intigration_tests/testcases.py
```diff
@@ -119,6 +119,11 @@ class Testcase:
         expected_output_xml_path="./expected_outputs/xml_profession_filter",
         expected_output_proto_path="./expected_outputs/proto_profession_filter",
     ),
+    Testcase(
+        name="animation_speed",
+        xml_input_paths=["./inputs/xml_animation_speed"],
+        expected_output_xml_path="./expected_outputs/xml_animation_speed",
+    ),
```
AsherGlick, line 126: Where are the proto testcases?

AsherGlick (changes requested):

xml_converter/intigration_tests/testcases.py
AsherGlick: Why not just change the proto here? Now seems like as good of a time as any to avoid it being forgotten.

AsherGlick (approved):

PR #230, opened 2023-11-16
Improve the bug workarounds section

Masgalor:
I think it is worth mentioning that KDE has proper tools to manage the window position without dragging it around after each start.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,4 @@ Known Bug Workarounds
 =====================
 * Burrito launches off-center or on the wrong monitor (This seems to be WM dependant.)
   * On Gnome and KDE it can be solved by holding down the super key and dragging burrito to the right position.
-  * On KDE it is also possible to use window-rules to set the position or a target monitor automatically. \
-    If you are not yet familiar with window-rules, you may find the examples in the official documentation useful. [Link](https://docs.kde.org/stable5/en/kwin/kcontrol/windowspecific/examples.html) \
-    To put it simply, you can add a new rule if you go to `System Settings → Window Behavior → Window Rules`. \
-    Remember to set `window class (application)` to `burrito`.
+  * On KDE it is also possible to use [window-rules](https://docs.kde.org/stable5/en/kwin/kcontrol/windowspecific/examples.html) to set the position or a target monitor automatically by adding a new rule to `System Settings → Window Behavior → Window Rules` and setting `window class (application)` to `burrito`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,5 +23,6 @@ Roadmap
 
 Known Bug Workarounds
 =====================
-* Burrito launches off-center or on the wrong monitor
-  * This seems to be WM dependant. on Gnome it can be solved by holding down the super key and dragging burrito to the right position.
+* Burrito launches off-center or on the wrong monitor (This seems to be WM dependant.)
+  * On Gnome and KDE it can be solved by holding down the super key and dragging burrito to the right position.
+  * On KDE it is also possible to use window-rules to set the position or a target monitor automatically.
```
AsherGlick, line 28: Can we insert a link to documentation or include some instructions as to how window-rules works? I want to make sure folks reading the docs either know what they need to know, or know where to look next.

Otherwise this looks good.

AsherGlick (changes requested):

README.md
Masgalor: Understood.
I didn't want to overdo it, because Gnome is the official target DE.
I've added some useful information about the KDE window management tools.

README.md
```diff
@@ -23,5 +23,9 @@ Roadmap
...
 Known Bug Workarounds
 =====================
-* Burrito launches off-center or on the wrong monitor
-  * This seems to be WM dependant. on Gnome it can be solved by holding down the super key and dragging burrito to the right position.
+* Burrito launches off-center or on the wrong monitor (This seems to be WM dependant.)
+  * On Gnome and KDE it can be solved by holding down the super key and dragging burrito to the right position.
+  * On KDE it is also possible to use window-rules to set the position or a target monitor automatically. \
+    If you are not yet familiar with window-rules, you may find the examples in the official documentation useful. [Link](https://docs.kde.org/stable5/en/kwin/kcontrol/windowspecific/examples.html) \
+    To put it simply, you can add a new rule if you go to `System Settings → Window Behavior → Window Rules`. \
+    Remember to set `window class (application)` to `burrito`.
```
AsherGlick, line 31: All the info looks like it is here. I think that this can be made more concise

```suggestion
  * On KDE it is also possible to use [window-rules](https://docs.kde.org/stable5/en/kwin/kcontrol/windowspecific/examples.html) to set the position or a target monitor automatically by adding a new rule to `System Settings → Window Behavior → Window Rules` and setting `window class (application)` to `burrito`.
```

AsherGlick (changes requested):

PR #231, opened 2023-11-18
Ensure that dir paths end in a forward slash when path is needed

klingbolt:
This ensures that when we are given a dir file path, whether it is `a/b` or `a/b/`, we are always using the trailing backslash in the strings.

### DIFF
--- a/xml_converter/src/packaging_protobin.cpp
+++ b/xml_converter/src/packaging_protobin.cpp
@@ -229,7 +229,7 @@ void write_protobuf_file_per_map_id(
     }
 
     for (auto iterator = mapid_to_category_to_pois.begin(); iterator != mapid_to_category_to_pois.end(); iterator++) {
-        string output_filepath = proto_directory + "/" + to_string(iterator->first) + ".data";
+        string output_filepath = proto_directory + to_string(iterator->first) + ".data";
 
         _write_protobuf_file(
             output_filepath,
--- a/xml_converter/src/string_helper.cpp
+++ b/xml_converter/src/string_helper.cpp
@@ -287,3 +287,10 @@ bool has_suffix(std::string const& fullString, std::string const& ending) {
         return false;
     }
 }
+
+void ensure_trailing_slash(std::string* directory_path) {
+    if (!has_suffix(*directory_path , "/")) {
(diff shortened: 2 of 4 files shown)


AsherGlick (changes requested):
This is a little confusing, breaking out the logic into its own function is fine. But then all these other changes are less "ensure the dir paths end in a forward slash" and more "lets just pray the dir paths end if a forward slash" because all of the sanitation checking was lifted out of the functions and into their callers.

EG: the first diff here: `write_protobuf_file_per_map_id` now cannot handle if you don't pass a string with a trailing slash. It just has to hope that whoever is calling it remembered to call "ensure_trailing_slash" beforehand.

xml_converter/src/string_helper.cpp
```diff
@@ -287,3 +287,10 @@ bool has_suffix(std::string const& fullString, std::string const& ending) {
         return false;
     }
 }
+
+void ensure_trailing_slash(std::string* directory_path) {
```
AsherGlick, line 291: As discussed offline, this is only used when you want to join two string paths together. This function should just do that instead needing to remove all the const references

AsherGlick (changes requested):

xml_converter/src/packaging_protobin.cpp
```diff
@@ -177,7 +177,7 @@ void _write_protobuf_file(
 }
 
 void write_protobuf_file(
-    const string& filepath,
+    string filepath,
```
AsherGlick, line 180: These should go back to being `const string&`s

AsherGlick (changes requested):

xml_converter/src/string_helper.cpp
```diff
@@ -287,3 +287,11 @@ bool has_suffix(std::string const& fullString, std::string const& ending) {
         return false;
     }
 }
+
+string join_file_paths(string directory_path, string file_path) {
```
AsherGlick, line 291: this can take in a `const string&` instead of a string too

AsherGlick (changes requested):

AsherGlick (approved):

PR #242, opened 2023-11-27
Dynamic minimap scale depending on ui-scale

Masgalor:
Depends on #225 
This PR aims to change the position and size of the minimap according to the used ui-scale.

The parameters `compassWidth` and `compassHeight` as read from the link only report values appropriate for the ui-scale `normal`. To get the actual size of the minimap we need to apply the same factor that is used to resize the menu buttons. Apart from that the only other thing that matters is a constant offset in the y-direction if the minimap is placed in the lower right corner.

### DIFF
--- a/Spatial.gd
+++ b/Spatial.gd
@@ -57,6 +57,12 @@ const button_margin = {
 	2: {"left": 361, "right": 394}, # large
 	3: {"left": 395, "right": 431}  # larger
 }
+const minimap_scale = {
+	0: {"offset": 32, "factor": 0.904}, # small
+	1: {"offset": 36, "factor": 1}, # normal
+	2: {"offset": 40, "factor": 1.118}, # large
+	3: {"offset": 44, "factor": 1.223} # larger
+}
 
 # Called when the node enters the scene tree for the first time.
 func _ready():
@@ -258,7 +264,7 @@ func decode_frame_packet(spb: StreamPeerBuffer):
 	if (!map_is_open):
 		map_size = Vector2(compass_width, compass_height)
 		if !compass_is_top_right:
-			map_corner = get_viewport().size - Vector2(compass_width, compass_height + 36)
+			map_corner = get_viewport().size - Vector2(compass_width, compass_height + self.minimap_scale[self.ui_size]["offset"])
 		else:
 			map_corner = Vector2(get_viewport().size.x - compass_width, 0)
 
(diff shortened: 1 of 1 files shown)


Spatial.gd
```diff
@@ -57,6 +57,12 @@ const button_margin = {
 	2: {"left": 361, "right": 394}, # large
 	3: {"left": 395, "right": 431}  # larger
 }
+const minimap_scale = {
+	0: {"offset": 32, "factor": 0.904}, # small
+	1: {"offset": 36, "factor": 1}, # normal
+	2: {"offset": 40, "factor": 1.118}, # large
+	3: {"offset": 44, "factor": 1.223} # larger
```
AsherGlick, line 64: Where did the `factor` values come from?

AsherGlick (commented):
Looks good in general. Just wondering what you used to come up with the values for `factor`

Spatial.gd
Masgalor: When doing the research for my other PR #225 I learned that the interface is always generated for the ui-scale `normal` and then rescaled before it gets drawn on the screen. I found that every single button and consequently the menu as a whole was rescaled with a constant factor with `normal` as base-value. So I used the values for the left margin of the burrito button to define those factors because this value also represents the total length of the main menu.

```
small: 292/323=0.904
normal: 323/323=1
large: 361/323=1.118
larger: 395/323=1.223
```

AsherGlick: Seems reasonable. But they seem so close to 10%/11% increments which would cause me to want to investigate further. Given that this pr is for the minimap, do you get the same four ratios when you count the pixels that the minimap takes up in each uisize? If it were me I would also want to check some of the non resizable windows and the skill bar too.

Masgalor: Every ui element is rescaled by the same factor.
After having a closer look I think it is safe to say that the scale is `90%` for `small`, `111%` for `large` and `122%` for `larger`. 

scale_minimap
scale_skillbar

Masgalor:
```suggestion
	0: {"offset": 32, "factor": 0.9}, # small
	1: {"offset": 36, "factor": 1}, # normal
	2: {"offset": 40, "factor": 1.11}, # large
	3: {"offset": 44, "factor": 1.22} # larger
```

AsherGlick: :white_check_mark:

AsherGlick (approved):

PR #251, opened 2023-12-30
Reorganizing Tests

AsherGlick:
Now that we have more of the tests and can see what their structure looks like the tests can be reorganized. This reorg combines all the info about a single test into a single place with a uniform structure with the other tests. Instead of having each test scattered across multiple locations. This also gets rid of the single source of all test info and splits up test info into each individual test.

Most of this PR is just moving files around, but there is also new code for loading the new tests, and a script for easily creating a new test with placeholder values.


AsherGlick:
Some of the build steps are failing. I am going to ignore them because they stem from moving the python directory up from `generators` to `xml_converter`. I'll fix them in the next PR to avoid our protodiff tool from creating a ton of messages. Probably the tool should become smarter to understand that it does not need to make a comment for empty diffs, and it should only show a diff for a file move if the original/target files have a diff between them.

klingbolt (approved):

Issue #80, opened 2022-07-25
[XML Markers] Call XML Parser Library from GDScript


AsherGlick:
Call the XML parser directly from gdscript without users needing to know the xml_parser exists separately. This will allow users to click the load / import button and import a marker pack from an XML file into the internal format, all without having to leave the burrito GUI.

PR #259, opened 2024-01-01
Calling xml converter

klingbolt:
Addressing #80 

This PR allows the user to select an xml pack and load the markers into the internal format. 

At this state, only one pack can be loaded at a time.


Spatial.tscn
```diff
@@ -863,24 +913,33 @@ color = Color( 0, 0, 0, 1 )
...
 mesh = SubResource( 3 )
 material/0 = SubResource( 4 )
 
+[node name="Control2" type="Control" parent="."]
+margin_right = 40.0
+margin_bottom = 40.0
+__meta__ = {
+"_edit_use_anchors_": false
+}
+
```
AsherGlick, line 930: What is this?

Spatial.tscn
```diff
@@ -863,24 +913,33 @@ color = Color( 0, 0, 0, 1 )
 [node name="Paths" type="Spatial" parent="."]
 
 [node name="Icons" type="Spatial" parent="."]
+transform = Transform( 1, 0, 0, 0, 1, 0, 0, 0, 1, 0.106085, 0, 0 )
```
AsherGlick, line 916: Why is this offset by `0.106085` now?

Spatial.tscn
```diff
@@ -183,20 +184,48 @@ __meta__ = {
...
+[node name="Label" type="Label" parent="Control/Dialogs/FileDialog/WindowDialog/ScrollContainer"]
+margin_right = 100.0
+margin_bottom = 100.0
+rect_min_size = Vector2( 100, 100 )
 __meta__ = {
 "_edit_use_anchors_": false
 }
 
 [node name="MainMenu" type="WindowDialog" parent="Control/Dialogs"]
+visible = true
```
AsherGlick, line 228: This should have been showing you a yellow warning that it should be hidden by default. (And would be hidden by default regardless of this setting)

Spatial.tscn
```diff
@@ -183,20 +184,48 @@ __meta__ = {
...
-margin_left = 636.0
-margin_top = 174.0
-margin_right = 1307.0
-margin_bottom = 672.0
-window_title = "Open a File"
-mode = 0
-access = 1
-current_dir = "user://"
-current_path = "user://"
+visible = true
```
AsherGlick, line 187: This should have been showing you a yellow warning that it should be hidden by default. (And would be hidden by default regardless of this setting)

Spatial.tscn
```diff
@@ -183,20 +184,48 @@ __meta__ = {
...
+margin_left = 312.0
+margin_top = 82.0
+margin_right = 983.0
+margin_bottom = 580.0
+window_title = "Open a Directory"
+mode = 2
+access = 2
+current_dir = "/home/steph/Desktop/Packs"
+current_path = "/home/steph/Desktop/Packs/"
```
AsherGlick, line 196: :(

Spatial.tscn
```diff
@@ -183,20 +184,48 @@ __meta__ = {
...
+script = ExtResource( 14 )
+__meta__ = {
+"_edit_use_anchors_": false
+}
+
+[node name="WindowDialog" type="WindowDialog" parent="Control/Dialogs/FileDialog"]
```
AsherGlick, line 202: Why is this window dialog inside of the file dialog?
Also why does it not have a name.
Also why does it always show up even if everything worked correctly

Spatial.tscn
```diff
@@ -183,20 +184,48 @@ __meta__ = {
...
+visible = true
```
AsherGlick, line 203: This should have been showing you a yellow warning that it should be hidden by default. (And would be hidden by default regardless of this setting)

Spatial.tscn
```diff
@@ -217,51 +246,72 @@ __meta__ = {
...
+margin_bottom = 72.0
+text = "Import Marker Pack"
+
+[node name="HSeparator6" type="HSeparator" parent="Control/Dialogs/MainMenu/ScrollContainer/VBoxContainer"]
+margin_top = 76.0
+margin_right = 220.0
+margin_bottom = 80.0
+__meta__ = {
+"_edit_use_anchors_": false
+}
```
AsherGlick, line 278: Lets delete this so that Import and Save are next to each other as they are logically similar.

Spatial.tscn
```diff
@@ -217,51 +246,72 @@ __meta__ = {
...
 [node name="PointEditor" type="Button" parent="Control/Dialogs/MainMenu/ScrollContainer/VBoxContainer"]
-visible = false
-margin_top = 96.0
-margin_right = 213.0
-margin_bottom = 136.0
+margin_top = 136.0
+margin_right = 220.0
+margin_bottom = 176.0
 rect_min_size = Vector2( 0, 40 )
 text = "Editor Panel"
```
AsherGlick, line 297: Is this duplicate "Editor Panel" button used anymore? It does not seem to have any signals attached to it.

Spatial.tscn
```diff
@@ -217,51 +246,72 @@ __meta__ = {
...
+margin_bottom = 48.0
+__meta__ = {
+"_edit_use_anchors_": false
+}
+
+[node name="ImportPath" type="Button" parent="Control/Dialogs/MainMenu/ScrollContainer/VBoxContainer"]
+margin_top = 52.0
+margin_right = 220.0
+margin_bottom = 72.0
+text = "Import Marker Pack"
```
AsherGlick, line 270: This button is a different size from the other buttons.

Spatial.gd
```diff
@@ -83,6 +84,8 @@ func _ready():
 
 	if (Settings.burrito_link_auto_launch_enabled):
 		launch_burrito_link()
+	
+
```
AsherGlick, line 88: ?

PackDialog.gd
```diff
@@ -0,0 +1,65 @@
...
+	self.user_data_dir = str(OS.get_user_data_dir())
+	self.protobin_data_folder = self.user_data_dir.plus_file("packs")
+	self.split_protobin_data_folder = self.user_data_dir.plus_file("protobins")
+	if not dir.dir_exists(self.protobin_data_folder):
+		var success = dir.make_dir(self.protobin_data_folder)
+		if success != OK:
+			print("Error: Could not create data folder:", self.protobin_data_folder)
+
+
+func clear_directory_contents(dir_path: String) -> void:
```
AsherGlick, line 21: This is **BY FAR** the most dangerous function you have ever written. Please add some sanity checks to it to make sure we are not recursively deleting the wrong directory/whole computer

PackDialog.gd
```diff
@@ -0,0 +1,65 @@
...
+		var success = dir.make_dir(new_path)
+		if success != OK:
+			print("Error: Could not create data folder:", self.protobin_data_folder)
+	#else: 
+	#	#Pop up here to confirm overwrite?
+	#	clear_directory_contents(dir_path)
+	var args: PoolStringArray = [
+		"--input-taco-path", dir_path,
+		"--output-waypoint-path", new_path,
+		"--output-split-waypoint-path", self.split_protobin_data_folder]
```
AsherGlick, line 51: nitpick:
```suggestion
		"--output-split-waypoint-path", self.split_protobin_data_folder
	]
```

AsherGlick (changes requested):
Code seems to crash on-load in the mistlock with the error
Screenshot from 2023-12-31 19-08-23

Is this caused by the same bug you mentioned was pre-existing? If so we can punt that fix to the next pr.

klingbolt:
> Code seems to crash on-load in the mistlock with the error Screenshot from 2023-12-31 19-08-23
> 
> Is this caused by the same bug you mentioned was pre-existing? If so we can punt that fix to the next pr.

No this looks like something else. Is the file being loaded newly created or an older export? It seems to suggest there is a trail field in the proto with no points.

AsherGlick:
I was able to trigger it to crash on startup when I had no converted markers, but I was also able to convert markers by not launching gw2 and using the burrito menus to convert markers. I converted the turtlepoint-skipbad markers and had the same issue.

I did have other old markers in the same burrito config directory but in a different subfolder. I attempted to remove them and was met with the same issue.

If there are no markers for a given map we probably should not crash, if that bug was introduced in this PR it should be fixed.

PackDialog.gd
klingbolt: Done

AsherGlick:
Screenshot from 2024-01-01 15-17-40
After discussing offline, we will ignore this mistlock issue as it is just bad data. We should not crash on bad data thoguh. However it seems that all marker  packs cannot be imported.

PackDialog.gd
klingbolt: I realize that and this should have not been kept in. I will delete.

Spatial.gd
klingbolt: Deleted.

Spatial.tscn
klingbolt: You are correct. Changing back to visible off.

Spatial.tscn
klingbolt: You are correct. Changing back to visible off.

Spatial.tscn
klingbolt: You are correct. Changing back to visible off.

Spatial.tscn
klingbolt: I missed adding the min size to it so it kept readjusting. Added.

Spatial.tscn
klingbolt: Sounds good. Done

Spatial.tscn
klingbolt: It is not. Removing.

Spatial.tscn
klingbolt: Not sure. I'm reverting this

Spatial.tscn
klingbolt: It was a test node. Removing.

PackDialog.gd
```diff
@@ -44,11 +32,11 @@ func _on_FileDialog_dir_selected(dir_path):
 			print("Error: Could not create data folder:", self.protobin_data_folder)
 	#else: 
 	#	#Pop up here to confirm overwrite?
-	#	clear_directory_contents(dir_path)
 	var args: PoolStringArray = [
 		"--input-taco-path", dir_path,
 		"--output-waypoint-path", new_path,
-		"--output-split-waypoint-path", self.split_protobin_data_folder]
+		"--output-split-waypoint-path", self.split_protobin_data_folder
+		]
```
AsherGlick: This should be outdented by one

Spatial.tscn
AsherGlick: This was not adressed

Spatial.gd
```diff
@@ -85,6 +86,7 @@ func _ready():
 		launch_burrito_link()
 
 
+
```
AsherGlick, line 89: ??

AsherGlick (changes requested):

Spatial.tscn
klingbolt: Removing this feature for now.

PackDialog.gd
```diff
@@ -37,15 +35,10 @@ func _on_FileDialog_dir_selected(dir_path):
 		"--output-waypoint-path", new_path,
 		"--output-split-waypoint-path", self.split_protobin_data_folder
 	]
-	custom_print(args)
-	var result: int = OS.execute(self.executable_path, args, true, output)
-	custom_print(output)
+	print(args)
+	var result: int = OS.execute(self.executable_path, args, true, output, true)
```
AsherGlick, line 39: Why set `open_console = true` ?

AsherGlick: Never-mind this is 4.2

AsherGlick (approved):

PR #261, opened 2024-01-03
Case Insensitive Category name

klingbolt:
Made the field name and type case insensitive. Added tests to ensure this.


xml_converter/integration_tests/test_cases/case_of_category_names/input/pack/xml_file.xml
```diff
@@ -0,0 +1,11 @@
+<OverlayData>
+    <MarkerCategory DisplayName="My Category" Name="MYCATEGORY">
+    </MarkerCategory>
+    <MarkerCategory DisplayName="My Category" Name="mycategory">
+    </MarkerCategory>
```
AsherGlick, line 5: Which of these two categories do the POIs belong to. Do we expect an error with having two categories using the same case insensitive name?

AsherGlick (changes requested):

xml_converter/integration_tests/test_cases/case_of_category_names/input/pack/xml_file.xml
klingbolt: If they have the same name, they should combine whenever we read them thus the lowercase is the one that is saved. This is standard xml procedure and I thought this would be a good place to ensure that is working correctly. This tests that the output for both xml and proto are what we expect in these cases.

AsherGlick: Lets put the second one in another file then to replicate the specific functionality you mention. While there are examples I can think of that would involve two marker category nodes of the same name in the same file to exist I don't necessarily want to encourage any of that behavior.

klingbolt: That makes more sense. Will update to that.

AsherGlick (approved):

Issue #27, opened 2021-08-15
Issue template?


coderedart:
people will definitely have issues with Burrito. as we can see in #25 where dev had to ask for distro/wm versions, it wastes time of everyone involved to be waiting for the basic info.

i think it would help if we can have an issue template guiding what info they are supposed to provide to help with debugging or pinpointing the issue. the most basic info they can provide is Versions of their Distro/ Desktop Environment or Window Manager / Lutris and the runtime used for gw2 / Burrito and burrito_link. 

and maybe the output/logs of burrito/burrito_link in terminal. 

 there seems to be a simple issue template from https://github.com/stevemao/github-issue-templates . 
also, we can provide the commands they need to run which they can copy paste to get the relevant versions. 
```
➜  ~ gnome-shell --version
GNOME Shell 40.3

➜  ~ cat /etc/os-release
NAME="EndeavourOS"
PRETTY_NAME="EndeavourOS"
ID=endeavouros
ID_LIKE=arch
BUILD_ID=2021.04.17

➜  ~ ./burrito.x86_64 --version
3.3.2.stable.official

```

AsherGlick:
This is a great idea. I have not used github issue templates before but they make a lot of sense.

PR #263, opened 2024-01-04
Create a form for bugreports

Masgalor:
I had a little time on my hands, so I've created a form for bugreports as requested in #27.

I wasn't sure how technically I should get with those questions so I decided to stay closer to the side of people which don't know that much about their system. However, just let me know what you want changed or added.

Unfortunately it is not easy to review issue-forms because they are not rendered correctly in the file preview.
You can test my current form in [this repo](https://github.com/Masgalor/issue-template/issues) or create an empty one yourself and test it there.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bugreport.yml
+++ b/.github/ISSUE_TEMPLATE/bugreport.yml
@@ -1,0 +1,57 @@
+name: Bug Report
+description: Something is crashing or not working as expected.
+title: "[Bug]: "
+projects: ["AsherGlick/Burrito"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this bug report!
+  - type: checkboxes
+    id: burrito_version
+    attributes:
+      label: Burrito Version
+      description: Please make sure you are using the latest version of Burrito before opening a report.
+      options:
+        - label: All components of Burrito (application, link and DLL) are upgraded to the latest version.
+          required: true
+  - type: dropdown
+    id: install_method
+    attributes:
+      label: Install
+      description: How did you install the link-component?
+      options:
+        - Standalone binary/script
+        - DLL (d3d11.dll)
+        - DLL (arcdps_burritolink.dll)
+        - Other
+      default: 0
+    validations:
+      required: true
+  - type: checkboxes
+    id: de_composing
+    attributes:
+      label: Desktop-Environment
+      description: Please mark everything that describes your system.
+      options:
+        - label: My desktop is Gnome
+        - label: My desktop is KDE
+        - label: I know for sure composing is enabled
+        - label: I'm using X11
+        - label: I'm using Wayland
+        - label: I have no idea what you are talking about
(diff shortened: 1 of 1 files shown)


AsherGlick:
I do think we eventually want something like this. I dont know what exactly it will look like so I will just share my general thoughts without necessarily providing guidance or solutions:

* Checkbox for "All components are upgraded" might be a good general idea but it is pretty hard for a user to confirm. burrito-link does not even really have a version number so the only way to guarantee this is by downloading whatever the newest version of the program is.
* I like the dropdown for asking about burrito-link. I might change it to reference "burrito link" specifically by name, and switch the ordering to have d3d11.dll on top as it is the recommended default. Even though the standalone version was first.
```yaml
  - type: dropdown
    id: burrito_link_install_method
    attributes:
      label: Burrito Link
      description: How did you install burrito link?
      options:
        - DLL (d3d11.dll)
        - DLL (arcdps_burritolink.dll)
        - Standalone (burrito_link.exe)
        - Other
      default: 0
```
* The desktop environment checkboxes are a little wonky. There is a lot of mixed info being presented there. They also translate into "tasks" in the issue itself, which lead to "2/7 tasks complete" for "My desktop is Gnome". The checkboxes are also limited in what the user can select. EG: AwesomeWM, XFCE, etc
* The `inxi -Gxxx && wmctrl -m` optional info is probably sufficient to replace those checkboxes. It is an ongoing project to determine even what info is necessary for debugging, you can find my incoherent ramblings https://github.com/AsherGlick/Burrito/wiki/Graphics-Stacks
* Github's Issue UI is not fantastic for "other" issues when templates exist. So we probably also want to include a template for feature requests, and maybe a template for questions, even if that template is just redirecting folks over to the discussions page.

AsherGlick:
> Checkbox for "All components are upgraded" might be a good general idea but it is pretty hard for a user to confirm. burrito-link does not even really have a version number so the only way to guarantee this is by downloading whatever the newest version of the program is.

Thinking about this more this question should just ask for the version number of Burrito/Burrito Link (once we figure out a good way to surface that info to the user) instead of a checkbox for "it is up to date". Because when we look back at old bugs we can see what version number those were submitted for instead of that they were using the most up to date version at the time the issue was submitted.

Masgalor:
>Thinking about this more this question should just ask for the version number of Burrito/Burrito Link (once we figure out a good way to surface that info to the user) instead of a checkbox for "it is up to date".

My idea with this checkbox was to give the user a hint to check for an update before opening a report.
I spent too much time of my life hunting bugs that didn't even exist anymore, because the customer reporting the bug used some ancient version. So I would like to keep the checkbox, not because it carries some useful information, but because it (hopefully) rises the awareness for software updates.
It makes sense to have the version in the report especially for long living issues, but there is no way for the user to know what version they are running, so this must be an optional question. I added an dropdown element for this, we need to remember to update it after a release.

I changed the form with your other mentions in mind, so have another look if you find the time.
I also added other forms for feature-requests and freestyle questions.

PR #264, opened 2024-01-05
Reworking copying functions to only copy files for textures

klingbolt:
The goal of this PR was to make it explicit when we want to copy images from an input folder to an output folder. Since this is adding copying files from the new protobin folders, I standardized how those are read to match the taco directories.


AsherGlick:
When would we have a situation where we have an output file and we dont want to copy files?

klingbolt:
We had previously mentioned if you are using the program simply as a linter, you would not want to copy them over. 

Perhaps we want to make that the optional case then and default to copying them.

AsherGlick:
And in that conversation we had also come to the conclusion that if you were using the xml_converter as only a linter you would not have an output file.

klingbolt:
Ah I had not remembered that. I'll make adjustments.

xml_converter/src/xml_converter.cpp
```diff
@@ -160,10 +183,24 @@ void process_data(
...
         cout << "Loading waypoint pack " << input_waypoint_paths[i] << endl;
-        read_protobuf_file(
+        read_burrito_directory(
             input_waypoint_paths[i],
             &marker_categories,
             &parsed_pois);
+
+        if (output_split_waypoint_dir != "") {
+            move_supplementary_files(input_waypoint_paths[i], output_split_waypoint_dir);
+        }
```
AsherGlick, line 193: This is going to copy more than just the necessary files. Which we only know about when we call the write data function.

xml_converter/src/xml_converter.cpp
```diff
@@ -160,10 +183,24 @@ void process_data(
...
+        if (output_taco_paths.size() != 0) {
+            for (size_t j = 0; j < output_taco_paths.size(); j++) {
+                move_supplementary_files(input_waypoint_paths[i], output_taco_paths[j]);
+            }
+        }
+        if (output_waypoint_paths.size() != 0) {
+            for (size_t j = 0; j < output_waypoint_paths.size(); j++) {
+                move_supplementary_files(input_waypoint_paths[i], output_waypoint_paths[j]);
+            }
+        }
```
AsherGlick, line 203: This is going to copy more than just the necessary files. Which we only know about when we call the write data function.

AsherGlick (changes requested):

klingbolt:
This change should now copy only the files that used as textures. In addition, the ability to read proto texture indexes was added.

AsherGlick (changes requested):

xml_converter/src/packaging_xml.cpp
```diff
@@ -160,14 +154,26 @@ void parse_xml_file(string xml_filepath, map<string, Category>* marker_categorie
 ////////////////////////////////////////////////////////////////////////////////
 ///////////////////////////////// DESERIALIZE //////////////////////////////////
 ////////////////////////////////////////////////////////////////////////////////
+void xml_post_processing(XMLWriterState* state) {
+    if (state->textures.size() > 1) {
```
AsherGlick, line 158: Why are you checking if there are two or more elements in the vector?

xml_converter/src/state_structs/xml_writer_state.hpp
```diff
@@ -1,7 +1,13 @@
...
+#include <vector>
+
+class Image;
 
 struct XMLWriterState {
-    std::string filedir;
+    const char* xml_filedir;
+
+    // A list of all of the textures with their paths.
+    std::vector<const Image*> textures;
```
AsherGlick, line 12: This datatype needs some more thought, it looks like you are just appending to it and then later iterating over it to copy everything out from it you might as well just copy stuff when you append to the list. I originally thought you were using this to do deduplication to prevent a texture that used by a hundred icons from being copied a hundred times.

AsherGlick (changes requested):
I reviewed part of this but it looks like there are some changes that are not totally necessary or thought out. I'm going to pass it back to you with only whatever notes I have so far so you can go over it again yourself first before sending it back again.

For example you changed a state variable from a string to a cstr and then because of that change you added a bunch of extra random function calls to deal with everything that cstr interacted with expecting a string.

Another is that you took the name `filedir` from the proto writer state struct and replaced other more descriptive names. I am not really sure what `filedir` is and probably that variable name should have been changed instead of being propagated throughout the codebase.

xml_converter/src/state_structs/xml_writer_state.hpp
```diff
@@ -2,6 +2,8 @@
 
 #include <string>
 
+class Image;
```
klingbolt, line 5: This should be removed

xml_converter/src/attribute/image.cpp
```diff
@@ -35,6 +37,11 @@ string image_to_xml_attribute(
     const string& attribute_name,
     XMLWriterState* state,
     const Image* value) {
+    if (filesystem::exists(filesystem::path(value->original_filepath))) {
+        filesystem::path output_path = filesystem::path(state->xml_directory) / value->filename;
+        filesystem::create_directories(output_path.parent_path());
+        filesystem::copy_file(filesystem::path(value->original_filepath), output_path, filesystem::copy_options::overwrite_existing);
+    }
```
AsherGlick, line 44: Lets add a warning if the the file does not exist.

xml_converter/src/packaging_protobin.cpp
```diff
@@ -154,7 +161,16 @@ void proto_post_processing(ProtoWriterState* state, waypoint::Waypoint* proto) {
...
+            const Image* image = state->textures[i];
+            texture_data->set_filepath(image->filename);
+            if (fs::exists(fs::path(state->textures[i]->original_filepath))) {
+                fs::path output_path = fs::path(state->proto_directory) / image->filename;
+                fs::create_directories(output_path.parent_path());
+                fs::copy_file(fs::path(image->original_filepath), output_path, fs::copy_options::overwrite_existing);
+            }
+            else {
+                cout << "Warning: File path " << state->textures[i]->original_filepath << " not found." << endl;
+            }
```
AsherGlick, line 173: As discussed offline, we want to move this file copy into image_to_proto's else statement which will have it only get executed the same number of times as this, but be in the same file as the xml image copy code.

xml_converter/src/packaging_protobin.cpp
```diff
@@ -185,25 +200,27 @@ void _write_protobuf_file(
...
-    proto_post_processing(&state, &output_message);
+    proto_post_processing(state, &output_message);
 
     output_message.SerializeToOstream(&outfile);
     outfile.close();
 }
 
 void write_protobuf_file(
-    const string& filepath,
+    const string& proto_directory,
```
AsherGlick, line 217: `proto_root_directory`

xml_converter/src/packaging_xml.cpp
```diff
@@ -160,14 +152,13 @@ void parse_xml_file(string xml_filepath, map<string, Category>* marker_categorie
 ////////////////////////////////////////////////////////////////////////////////
 ///////////////////////////////// DESERIALIZE //////////////////////////////////
 ////////////////////////////////////////////////////////////////////////////////
-
-void write_xml_file(string xml_filepath, map<string, Category>* marker_categories, vector<Parseable*>* parsed_pois) {
+void write_xml_file(const string xml_directory, map<string, Category>* marker_categories, vector<Parseable*>* parsed_pois) {
```
AsherGlick, line 155: xml_directory -> xml_root_directory

xml_converter/src/state_structs/xml_reader_state.hpp
```diff
@@ -1,11 +1,9 @@
 #pragma once
 
-#include <map>
 #include <string>
 
 class Category;
 
 struct XMLReaderState {
-    std::string xml_filedir;
-    std::map<std::string, Category>* marker_categories;
```
AsherGlick, line 10: I think we need this data to do some validations against some fields that reference categories, even if we are not using it right now.

xml_converter/src/xml_converter.cpp
```diff
@@ -62,57 +62,69 @@ vector<string> get_files_by_suffix(string directory, string suffix) {
...
+void write_taco_directory(
+    string output_path,
+    map<string, Category>* marker_categories,
+    vector<Parseable*>* parsed_pois) {
     // TODO: Exportion of XML Marker Packs File Structure #111
     if (!filesystem::is_directory(output_path)) {
         if (!filesystem::create_directory(output_path)) {
             cout << "Error: " << output_path << "is not a valid directory path" << endl;
         }
+        return;
```
AsherGlick, line 110: I think this is in the wrong scope, it should be in the if statement above it for when a directory is not created properly. Otherwise this will always exit even if a directory was created properly.

AsherGlick (changes requested):

AsherGlick (approved):

PR #267, opened 2024-01-20
Save current state to split files

klingbolt:
Changes to markers are saved in the split files.

The next PR will save the splits to individual packs.

### DIFF
--- a/Gizmo/PointEdit.gd
+++ b/Gizmo/PointEdit.gd
@@ -29,16 +29,6 @@ func select(camera, event):
 	emit_signal("selected", self)
 
 
-func link_point(point_type: String, object_link, object_2d_link = null, object_index = 0):
-	self.point_type = point_type
-	self.object_link = object_link
-	if point_type == "path" || point_type == "area":
-		self.object_2d_link = object_2d_link
-		self.object_index = object_index
-	if point_type == "icon":
-		pass
-
-
 func update_point():
 	if self.translation != self.last_translation:
 		print("update")
--- a/Gizmo/PointEdit.gd
+++ b/Gizmo/PointEdit.gd
@@ -3,7 +3,7 @@ extends Spatial
 var camera: Camera
 signal selected(selected_object)
 signal deselected(selected_object)
-signal update()
+signal update(point_position)
 var last_translation
 var selected: bool = false
 
@@ -28,11 +28,6 @@ func select(camera, event):
 	$Pillar/CollisionShape.disabled = false
 	emit_signal("selected", self)
 
-var object_link = null
-var object_2d_link = null
-var object_index:int = 0
-var point_type: String
-
 
 func link_point(point_type: String, object_link, object_2d_link = null, object_index = 0):
 	self.point_type = point_type
@@ -46,14 +41,9 @@ func link_point(point_type: String, object_link, object_2d_link = null, object_i
 
 func update_point():
 	if self.translation != self.last_translation:
-		if point_type == "path" || point_type == "area":
-			self.object_link.set_point_position(self.object_index, self.translation)
-			self.object_2d_link.points[self.object_index] = Vector2(self.translation.x, self.translation.z)
-		if point_type == "icon":
-			self.object_link.set_point_position(self.translation)
 		print("update")
-		emit_signal("update")
-		self.last_translation  = self.translation
+		emit_signal("update", self.translation)
+		self.last_translation = self.translation
 
 ################################################################################
 # Handle resizing the control nodes so that no matter how far away from the
--- a/Gizmo/PointEdit.gd
+++ b/Gizmo/PointEdit.gd
@@ -3,6 +3,7 @@ extends Spatial
 var camera: Camera
 signal selected(selected_object)
 signal deselected(selected_object)
+signal update()
 var last_translation
 var selected: bool = false
 
@@ -49,10 +50,10 @@ func update_point():
 			self.object_link.set_point_position(self.object_index, self.translation)
 			self.object_2d_link.points[self.object_index] = Vector2(self.translation.x, self.translation.z)
 		if point_type == "icon":
-			self.object_link.translation = self.translation
+			self.object_link.set_point_position(self.translation)
 		print("update")
+		emit_signal("update")
 		self.last_translation  = self.translation
-
 
 ################################################################################
 # Handle resizing the control nodes so that no matter how far away from the
--- a/Icon.gd
+++ b/Icon.gd
@@ -6,6 +6,12 @@ var texture_path
 var waypoint: Waypoint.Icon
 var category: TreeItem
 
+func update_waypoint_icon():
+	var position = self.waypoint.position()
+	position.set_x(translation[0])
+	position.set_y(translation[1])
+	position.set_z(-translation[2])
+
 func set_icon_image(texture_path: String):
 	self.texture_path = texture_path
 	
--- a/Icon.gd
+++ b/Icon.gd
@@ -8,7 +8,7 @@ var category: TreeItem
 
 func set_point_position(point_position: Vector3):
 	self.translation = point_position
-	var position = self.waypoint.position()
+	var position = self.waypoint.get_position()
 	if position == null:
 		position = Waypoint.Position.new()
 	position.set_x(point_position[0])
--- a/Icon.gd
+++ b/Icon.gd
@@ -6,11 +6,14 @@ var texture_path
 var waypoint: Waypoint.Icon
 var category: TreeItem
 
-func update_waypoint_icon():
+func set_point_position(point_position: Vector3):
+	self.translation = point_position
 	var position = self.waypoint.position()
-	position.set_x(translation[0])
-	position.set_y(translation[1])
-	position.set_z(-translation[2])
+	if position == null:
+		position = Waypoint.Position.new()
+	position.set_x(point_position[0])
+	position.set_y(point_position[1])
+	position.set_z(-point_position[2])
 
 func set_icon_image(texture_path: String):
 	self.texture_path = texture_path
--- a/Spatial.gd
+++ b/Spatial.gd
@@ -9,7 +9,7 @@ var map_is_open: bool
 var compass_is_top_right: bool
 
 var edit_panel_open: bool = false
-var unsaved_changes: bool = false
+var unsaved_changes: bool = false setget set_unsaved_changes
 
 # This is the path to the texture that will be used for the next created 3d-path
 # object or icon object in the UI
@@ -55,6 +55,7 @@ const gizmo_scene = preload("res://Gizmo/PointEdit.tscn")
 const CategoryData = preload("res://CategoryData.gd")
 const Waypoint = preload("res://waypoint.gd")
 const PackDialog = preload("res://PackDialog.gd")
+const SelectedNode = preload("res://SelectedNode.gd")
 
 ##########Node Connections###########
 onready var markers_ui := $Control/Dialogs/CategoriesDialog/MarkersUI as Tree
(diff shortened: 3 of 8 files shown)


Spatial.gd
```diff
@@ -632,31 +638,18 @@ func gen_new_icon(position: Vector3, texture_path: String, waypoint_icon, catego
 	var category_data = category_item.get_metadata(0)
 	category_data.category3d.add_icon(new_icon)
 
-# This function take all of the currently rendered objects and converts it into
-# the data format that is saved/loaded from.
-func data_from_renderview():
```
AsherGlick, line 637: If we are no longer relying on the internal state of the nodes how can we be sure that the state of the nodes is in sync with whatever we are saving

Spatial.gd
```diff
@@ -632,31 +638,18 @@ func gen_new_icon(position: Vector3, texture_path: String, waypoint_icon, catego
...
-		paths_data.append({
-			"points": points,
-			"texture": path.texture_path
-		})
-
-	var data_out = {"icons": icons_data, "paths": paths_data}
-	return data_out
+func save_current_map_data():
+	var packed_bytes = self.waypoint_data.to_bytes()
+	if packed_bytes.size() > 0:
```
AsherGlick, line 649: Does this check prevent saving a file which went from having "some markers" to "no markers" ?
IE: prevent deleting the last marker on a map?

Spatial.gd
```diff
@@ -822,16 +815,14 @@ func _on_NewPathPoint_pressed():
...
 
 ################################################################################
 # TODO: This function will be used when exporting packs
 ################################################################################
 func _on_SaveDialog_file_selected(path):
-	self.markerdata[str(self.map_id)] = data_from_renderview()
-	var save_game = File.new()
-	save_game.open(path, File.WRITE)
-	save_game.store_string(JSON.print(self.markerdata))
+	pass
```
AsherGlick, line 825: If this function is not doing anything is is needed anymore?

Spatial.gd
```diff
@@ -905,6 +896,8 @@ func _on_ReversePathDirection_pressed():
 
 
 func _on_ExitButton_pressed():
+	if unsaved_data_icon.visible:
```
AsherGlick, line 899: ditto about

> This is a very weird highly coupled check on if an effectively random node happens to be visible or not. This check should be a more direct "if there are changes on the map then save" not "if there is a visible ui element that is telling the user there are changes on a map then save"

Spatial.tscn
```diff
@@ -184,10 +195,10 @@ __meta__ = {
...
 
 [node name="ImportPackDialog" type="FileDialog" parent="Control/Dialogs"]
-margin_left = 289.0
-margin_top = 36.0
-margin_right = 960.0
-margin_bottom = 534.0
+margin_left = 777.0
+margin_top = 89.0
+margin_right = 1448.0
+margin_bottom = 587.0
```
AsherGlick, line 201: What is the motivation for moving the window?

Spatial.tscn
```diff
@@ -94,6 +94,17 @@ __meta__ = {
...
+visible = false
+modulate = Color( 0.92549, 0, 0, 0.439216 )
+margin_top = 4.0
+margin_right = 25.0
+margin_bottom = 29.0
+texture = ExtResource( 3 )
+__meta__ = {
+"_edit_use_anchors_": false
+}
+
```
AsherGlick, line 107: Do we need an entire separate node with its own offsets and details if all we are doing is modulating the color?
It should be decided if the correct behavior is color modulation, or a notification circle, or w/e.

This comment questioning how this feature should work should also go to show why checking if this node is visible or not to save/not-save a marker pack is unnecessary coupling. If all these changes went through and then later it was decided to just get rid of this node and modulate the base button then all of burrito would be unable to save for no reason.

AsherGlick (changes requested):

Icon.gd
```diff
@@ -6,6 +6,12 @@ var texture_path
 var waypoint: Waypoint.Icon
 var category: TreeItem
 
+func update_waypoint_icon():
+	var position = self.waypoint.position()
+	position.set_x(translation[0])
+	position.set_y(translation[1])
+	position.set_z(-translation[2])
```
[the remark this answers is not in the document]
klingbolt: I've edited this so that it can now set a new value or overwrite an existing value

Spatial.gd
klingbolt: We should be relying on the proto as the source of truth. So Icon.gd and Route.gd should now be set to update the proto any time a change is made to the node.

Spatial.gd
klingbolt: I was using the check to not make a blank file but you are correct it would prevent that. I'm going to remove the check.

Spatial.gd
klingbolt: You're correct. I'll remove it.

Spatial.tscn
klingbolt: I needed a clear indicator when i was doing testing. After looking at it, I'm removing this node and just modulating the existing icon.

Spatial.tscn
klingbolt: There are several windows that overlap each other. This moves this window to an unused space. This is something to keep in mind when we do the larger UI update.

Spatial.gd
AsherGlick: Make and link issue to track this theoretical case of creating blank files. Because that is possible, maybe the correct logic here will be to delete files that are empty instead of not saving/overwriting them. But something a future pass can deal with.

Spatial.gd
```diff
@@ -641,15 +641,21 @@ func gen_new_icon(position: Vector3, texture_path: String, waypoint_icon, catego
...
 	var packed_bytes = self.waypoint_data.to_bytes()
-	if packed_bytes.size() > 0:
-		var file = File.new()
-		file.open(self.marker_file_path, file.WRITE)
-		file.store_buffer(packed_bytes)
+	var file = File.new()
+	file.open(self.marker_file_path, file.WRITE)
+	file.store_buffer(packed_bytes)
+	self.unsaved_changes = false
+	update_burrito_icon()
```
AsherGlick, line 658: How easy/likely is it for a programming mistake to lead to a desynced status and saved notification? EG: updating the variable without calling update_burrito_icon()

AsherGlick (changes requested):

Spatial.gd
```diff
@@ -641,15 +641,21 @@ func gen_new_icon(position: Vector3, texture_path: String, waypoint_icon, catego
...
 # Section of functions for saving changes to markers
 ################################################################################
-func on_change_made():
-	self.unsaved_data_icon.visible = true
+func update_burrito_icon():
+	if self.unsaved_changes:
+		$Control/GlobalMenuButton/TextureRect.modulate = Color(.78, 0.3, 0.3, 1.0)
+		$Control/GlobalMenuButton/main_menu_toggle.hint_tooltip = "Unsaved Data"
+	else:
+		$Control/GlobalMenuButton/TextureRect.modulate = Color(1.0, 1.0, 1.0, 1.0)
```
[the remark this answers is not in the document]
klingbolt: Understood. Using the white color constant. I also found that the alpha should have been lower than one to match the original value.

Gizmo/PointEdit.gd
```diff
@@ -49,11 +50,11 @@ func update_point():
 			self.object_link.set_point_position(self.object_index, self.translation)
 			self.object_2d_link.points[self.object_index] = Vector2(self.translation.x, self.translation.z)
 		if point_type == "icon":
-			self.object_link.translation = self.translation
+			self.object_link.set_point_position(self.translation)
```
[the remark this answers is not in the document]
klingbolt: Per our discussion, I have moved the functionality here into the main file. Gizmo should no longer care what type the point is put instead just emit its position when there is a change. This change did require me to make a class that can be used to contain all of the information gizmo was holding.

Spatial.gd
```diff
@@ -388,13 +388,13 @@ func reset_3D_minimap_masks(category: Spatial):
 		reset_3D_minimap_masks(subcategory)
 
 
-var waypoint_data = Waypoint.Waypoint.new()
-var split_marker_file_dir = "user://protobins/"
-var marker_pack_dir = "user://packs/"
-var marker_file_path = ""
+var waypoint_data: Waypoint.Waypoint = Waypoint.Waypoint.new()
+const user_protobin_by_map_id_dir: String = "user://protobin_by_map_id/"
+const user_marker_pack_dir: String = "user://marker_packs/"
```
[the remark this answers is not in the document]
klingbolt: I like `downloaded_marker_packs_dir` for the 2nd one.

How about something like `active_data_dir`. Any changes made with the editor would be saved in this folder and "active_data" conveys that it is mutable.

Spatial.tscn
```diff
@@ -85,7 +85,7 @@ __meta__ = {
 }
 
 [node name="TextureRect" type="TextureRect" parent="Control/GlobalMenuButton"]
-modulate = Color( 1, 1, 1, 0.439216 )
+modulate = Color( 1, 1, 1, 0.44 )
```
klingbolt, line 88: Rounded this to make it simpler to reset.

Spatial.tscn
```diff
@@ -185,7 +185,7 @@ __meta__ = {
 
 [node name="ImportPackDialog" type="FileDialog" parent="Control/Dialogs"]
 margin_left = 289.0
-margin_top = 36.0
+margin_top = 60.0
```
klingbolt, line 188: Moved this back but shrunk it down a bit so it didn't overlap the editor bar and burrito icon

Spatial.gd
```diff
@@ -690,30 +700,35 @@ func gen_adjustment_nodes():
 				continue
 			var new_gizmo = gizmo_scene.instance()
 			new_gizmo.translation = gizmo_position
-			new_gizmo.link_point("path", route, path2d, i)
-			new_gizmo.connect("selected", self, "on_gizmo_selected")
+			new_gizmo.connect("selected", self, "on_gizmo_selected", ["path", [route, path2d], i])
```
AsherGlick, line 703: I think something is too clever here passing in an array of [route, path2d] or [icon] depending on the first argument being "path" or "icon".

Spatial.gd
```diff
@@ -690,30 +700,35 @@ func gen_adjustment_nodes():
...
-func on_gizmo_selected(object):
-	self.currently_selected_node = object
+func on_gizmo_selected(object, point_type: String, nodes: Array, point_index: int):
+	self.currently_selected_node = SelectedNode.new()
+	self.currently_selected_node.gizmo = object
+	self.currently_selected_node.point_type = point_type
 	$Control/Dialogs/NodeEditorDialog/ScrollContainer/VBoxContainer/DeleteNode.disabled = false
 	# Only enable these buttons if the object selected is a point on the path not an icon
-	if object.point_type == "path":
+	if point_type == "path":
```
AsherGlick, line 723: Can we have logic that does not rely on `point_type` branching in multiple places?

Spatial.gd
```diff
@@ -729,9 +744,14 @@ func on_gizmo_deselected(object):
 	$Control/Dialogs/NodeEditorDialog/ScrollContainer/VBoxContainer/SetActivePath.disabled = true
 	$Control/Dialogs/NodeEditorDialog/ScrollContainer/VBoxContainer/ReversePathDirection.disabled = true
 
-func on_update_made():
-	self.unsaved_changes = true
-	update_burrito_icon()
+func on_point_updated(position: Vector3):
+	set_unsaved_changes(true)
+	if self.currently_selected_node.point_type == "icon":
```
AsherGlick, line 749: Can we have logic that does not rely on `point_type` branching in multiple places?

Spatial.gd
```diff
@@ -833,28 +854,28 @@ func _on_NodeEditorDialog_hide():
 
 func _on_DeleteNode_pressed():
 	if self.currently_selected_node.point_type == "icon":
```
AsherGlick, line 856: Can we have logic that does not rely on `point_type` branching in multiple places?

Spatial.gd
```diff
@@ -833,28 +854,28 @@ func _on_NodeEditorDialog_hide():
...
 		path.remove_point(index)
 		path2d.remove_point(index)
 	clear_adjustment_nodes()
 	gen_adjustment_nodes()
-	on_gizmo_deselected(self.currently_selected_node)
+	on_gizmo_deselected(self.currently_selected_node.gizmo)
 
 
 func _on_NewNodeAfter_pressed():
 	if self.currently_selected_node.point_type == "icon":
```
AsherGlick, line 872: Can we have logic that does not rely on `point_type` branching in multiple places?

AsherGlick (changes requested):
As we discussed offline, this might actually be 3 PRs
1) Removing the tight coupling of 3d and 2d routes in gizmo
2) Changing the source of truth of the node positions from 3d nodes to waypoint datums
3) Saving the waypoint data to file

Lets do a technical write up on number 2 too before continuing.

klingbolt:
This PR is now obsolete.

PR #268, opened 2024-01-20
Godot File Handler

klingbolt:
Changed name of file that will hold all functions for loading and saving files. Also deleted the obsolete JSON code.

### DIFF
--- a/FileHandler.gd
+++ b/FileHandler.gd
@@ -1,16 +1,4 @@
-var downloaded_markers_dir: String
-var unsaved_markers_dir: String
-var user_data_dir: String 
-
-func _init():
-	self.user_data_dir = str(OS.get_user_data_dir())
-	self.downloaded_markers_dir = self.user_data_dir.plus_file("packs")
-	self.unsaved_markers_dir = self.user_data_dir.plus_file("protobins")
-	create_directory_if_missing(self.downloaded_markers_dir)
-	create_directory_if_missing(self.unsaved_markers_dir)
-
-static func call_xml_converter(args):
-	var executable_path: String = "./xml_converter/build/xml_converter"
+static func call_xml_converter(executable_path: String, args: PoolStringArray):
 	var output: Array = []
 	print(args)
 	var result: int = OS.execute(executable_path, args, true, output, true)
@@ -20,7 +8,7 @@ static func call_xml_converter(args):
 	else:
 		print("Command executed successfully.")
 
-static func create_directory_if_missing(path):
+static func create_directory_if_missing(path: String):
 	var dir = Directory.new()
 	if not dir.dir_exists(path):
 		var success = dir.make_dir(path)
--- a/FileHandler.gd
+++ b/FileHandler.gd
@@ -6,35 +6,14 @@ var split_protobin_data_folder: String
 var user_data_dir: String 
 
 func _ready():
-	var dir = Directory.new()
 	self.user_data_dir = str(OS.get_user_data_dir())
 	self.protobin_data_folder = self.user_data_dir.plus_file("packs")
 	self.split_protobin_data_folder = self.user_data_dir.plus_file("protobins")
-	if not dir.dir_exists(self.protobin_data_folder):
-		var success = dir.make_dir(self.protobin_data_folder)
-		if success != OK:
-			print("Error: Could not create data folder:", self.protobin_data_folder)
-	if not dir.dir_exists(self.split_protobin_data_folder):
-		var success = dir.make_dir(self.split_protobin_data_folder)
-		if success != OK:
-			print("Error: Could not create data folder:", self.split_protobin_data_folder)
+	ensure_folder_existance(self.protobin_data_folder)
+	ensure_folder_existance(self.split_protobin_data_folder)
 
-func _on_FileDialog_dir_selected(dir_path):
+func call_xml_converter(args):
 	var output: Array = []
-	print("Selected folder:", dir_path)
-	var dir = Directory.new()
-	var new_path: String = self.protobin_data_folder.plus_file(dir_path.get_file())
-	if not dir.dir_exists(new_path):
-		var success = dir.make_dir(new_path)
-		if success != OK:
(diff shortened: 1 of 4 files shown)


Spatial.gd
```diff
@@ -51,9 +51,14 @@ const category3d_scene = preload("res://Category3D.tscn")
...
 const gizmo_scene = preload("res://Gizmo/PointEdit.tscn")
+
+# Scripts containing code used by this scene
 const CategoryData = preload("res://CategoryData.gd")
 const Waypoint = preload("res://waypoint.gd")
-const PackDialog = preload("res://PackDialog.gd")
+const FileHandler = preload("res://FileHandler.gd")
+
+# Instancing scripts
+var file_handler: FileHandler = FileHandler.new()
```
AsherGlick, line 61: This is never used

Spatial.gd
```diff
@@ -80,6 +85,9 @@ func _ready():
 	OS.set_window_position(Vector2(0,0))
 	set_minimal_mouse_block()
 
+	# Call ready for additional scripts
+	self.file_handler._ready()
```
AsherGlick, line 89: Why are you manually calling `_ready()` which is a function that is called automatically?

Spatial.gd
```diff
@@ -822,16 +804,13 @@ func _on_NewPathPoint_pressed():
...
 
 ################################################################################
 # TODO: This function will be used when exporting packs
 ################################################################################
 func _on_SaveDialog_file_selected(path):
-	self.markerdata[str(self.map_id)] = data_from_renderview()
-	var save_game = File.new()
-	save_game.open(path, File.WRITE)
-	save_game.store_string(JSON.print(self.markerdata))
+	pass
```
AsherGlick, line 813: Are these functions no longer needed? If so then they can just be removed.

Spatial.gd
```diff
@@ -632,32 +640,6 @@ func gen_new_icon(position: Vector3, texture_path: String, waypoint_icon, catego
...
-			var point_position:Vector3 = path.get_point_position(point)
-			points.append([point_position.x, point_position.y, -point_position.z])
-		paths_data.append({
-			"points": points,
-			"texture": path.texture_path
-		})
-
-	var data_out = {"icons": icons_data, "paths": paths_data}
-	return data_out
-
```
AsherGlick, line 660: If we are no longer gathering this information from the visible nodes, what other mechanisms are in place to make sure that the data does not drift between the visual and whatever data we are saving?

AsherGlick (changes requested):
This PR seems to remove all capability of saving, is that desirable? Was saving anything already fully broken, not just doing the wrong thing, and this is just a cleanup PR?

klingbolt:
> This PR seems to remove all capability of saving, is that desirable? Was saving anything already fully broken, not just doing the wrong thing, and this is just a cleanup PR?

Everything was broken for saving. The nodes it was calling from never had anything written to it.

ImportPackDialog.gd
```diff
@@ -0,0 +1,20 @@
+extends FileDialog
+const FileHandler = preload("res://FileHandler.gd")
```
AsherGlick, line 2: FileHandler should probably be a static class which extends nothing if it is just being used like a library. This means it does not need to be initialized with `.new()` either.

AsherGlick (changes requested):

FileHandler.gd
```diff
@@ -1,28 +1,26 @@
...
-	ensure_folder_existance(self.protobin_data_folder)
-	ensure_folder_existance(self.split_protobin_data_folder)
+	self.downloaded_markers_dir = self.user_data_dir.plus_file("packs")
+	self.unsaved_markers_dir = self.user_data_dir.plus_file("protobins")
+	create_directory_if_missing(self.downloaded_markers_dir)
+	create_directory_if_missing(self.unsaved_markers_dir)
 
-func call_xml_converter(args):
+static func call_xml_converter(args):
+	var executable_path: String = "./xml_converter/build/xml_converter"
```
AsherGlick, line 13: This should go back to where it used to be, no reason to recreate the variable each time this function is called.

ImportPackDialog.gd
```diff
@@ -1,20 +1,15 @@
...
-
-func _ready():
-	self.file_handler = FileHandler.new()
-	pass
 
 func _on_FileDialog_dir_selected(dir_path):
 	print("Selected folder:", dir_path)
-	var new_path: String = file_handler.protobin_data_folder.plus_file(dir_path.get_file())
-	file_handler.ensure_folder_existance(new_path)
+	var new_path: String = FileHandler.downloaded_markers_dir.plus_file(dir_path.get_file())
```
AsherGlick, line 6: As discussed offline, this crashes. Please be sure to test the code before submitting it.

ImportPackDialog.gd
```diff
@@ -1,20 +1,15 @@
...
+	FileHandler.create_directory_if_missing(new_path)
 	#else: 
 	#	#Pop up here to confirm overwrite?
 	var args: PoolStringArray = [
 		"--input-taco-path", dir_path,
 		"--output-waypoint-path", new_path,
-		"--output-split-waypoint-path", file_handler.split_protobin_data_folder
+		"--output-split-waypoint-path", FileHandler.unsaved_markers_dir
```
AsherGlick, line 13: This will probably crash too for the same reason as above.

AsherGlick (changes requested):

FileHandler.gd
```diff
@@ -0,0 +1,16 @@
+static func call_xml_converter(args: PoolStringArray):
+	var output: Array = []
+	print(args)
+	var result: int = OS.execute("./xml_converter/build/xml_converter", args, true, output, true)
```
AsherGlick, line 4: This can probably be a constant variable in FileHandler instead of a magic string here.

Also: does this work with the exported files? I think it is fine if not right now but we will need to test it to be sure.

AsherGlick (changes requested):

AsherGlick (approved):

PR #289, opened 2024-05-19
Achievement Bit fixes

AsherGlick:
We had previously assumed that achievement bit was a bitmask, but it is an index.
This PR fixes this misconception and puts effort to prevent it from occuring again by
* Renaming the proto filed to bit_index
* Adding an xml field option of `AchievementBitIndex` as an alternative to `AchievementBit`
* Removing or renaming anything that used `achievement bitmask`

This PR will have wonky diffs, the best way to review this is to review it one commit at a time. The first commit is the real diff, the second commit is just renaming files.

Additionally the protobin diffs are going to be very strange. This is for two reasons
* The proto diff code does not use the most up-to-date version of the protodef because it runs on the branch we are trying to merge into instead of the branch we are trying to merge
* I committed a cardinal sin of re-indexing the proto fields. 16 used to be achievement bit, but now it is achievement. 17 used to be achievement but is now achievement_bit
* That cardinal sin is fine because we still have not released this proto. Anyone who is using it already should be ready to suffer the consequences of it changing like this.


xml_converter/integration_tests/test_cases/achievement_bitmask_valid/input/pack/xml_file.xml
```diff
@@ -7,13 +7,6 @@
...
 		<Trail AchievementBit="5" Type="mycategory" />
-
-		<!-- TODO: textproto seems to think a fixed32 is a uint32, but we support negative values for this currently. We should probably adjust that -->
-		<POI AchievementBit="-2147483648" Type="mycategory" XPos="169.81" YPos="210.65" ZPos="215.83" MapID="50" />
-		<POI AchievementBit="-5" Type="mycategory" XPos="169.81" YPos="210.65" ZPos="215.83" MapID="50" />
-
-		<!-- TODO: These throw errors and crash but probably should not if this were a uint32 -->
-		<!-- <POI AchievementBit="4294967295" Type="mycategory" XPos="169.81" YPos="210.65" ZPos="215.83" MapID="50" /> -->
-		<!-- <POI AchievementBit="2147483648" Type="mycategory" XPos="169.81" YPos="210.65" ZPos="215.83" MapID="50" /> -->
+		<POI AchievementBit="2147483647" Type="mycategory" XPos="169.81" YPos="210.65" ZPos="215.83" MapID="50" />
```
klingbolt: This line is identical to line 8 now. Is this intended?

klingbolt (changes requested):

xml_converter/integration_tests/test_cases/achievement_bitmask_valid/input/pack/xml_file.xml
AsherGlick: Good catch, removed.

klingbolt (approved):

PR #294, opened 2024-05-20
setting up a parallel method of accessing metadata contents

AsherGlick:
The goal will be to eventually completely replace the existing methods for accessing the metadata with this fully typed way to make the code faster to write and easier to read. This will ultimately remove JSONSchema from our list of dependencies once complete.


AsherGlick:
These tests are failing because I have used a feature of dataclasses `kw_only=True`. This feature only exists in python 3.10 and the version of python Ubuntu 20.04 is not 3.10. Looking to see what options are available here.

klingbolt (approved):

PR #296, opened 2024-05-20
Using signals to update the 2d line when changes are made to the 3d line

klingbolt:
This removes the need to have a linked path2d in the gizmo as any change to the point list in the route can be signaled to change in the 2d path as well.

### DIFF
--- a/Gizmo/PointEdit.gd
+++ b/Gizmo/PointEdit.gd
@@ -28,16 +28,14 @@ func select(camera, event):
 	emit_signal("selected", self)
 
 var object_link = null
-var object_2d_link = null
 var object_index:int = 0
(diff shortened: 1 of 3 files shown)


AsherGlick:
We discussed this offline but I realized this morning I am going to forget unless I make a note here. I think it is not a great idea for the 2d and 3d nodes to be this tightly coupled. EG: In order to update the 2d node you update the 3d node and have that information propagate to the 2d node. One of the things we discussed several times is how these different pieces of code should be more loosely coupled. Ironically the reason many of these other changes even need to be made is because gizmo was/is so tightly coupled with these two types, so you can see what dangers having highly coupled code brings.

PR #297, opened 2024-05-21
Migrating the cpp generator to use dataclasses

AsherGlick:
This migrates the rest of the code that uses the markdown frontmatter metadata to use dataclasses.

### DIFF
--- a/xml_converter/generators/generate_cpp.py
+++ b/xml_converter/generators/generate_cpp.py
@@ -1,8 +1,9 @@
 from dataclasses import dataclass, field
-from typing import Set, List, Dict, Optional, Tuple, TypedDict, Any
+from typing import Set, List, Dict, Optional, Tuple, TypedDict
 from jinja2 import FileSystemLoader, Environment, Template
+from metadata import CustomFunction, MetadataType
 from jinja_helpers import UnindentBlocks
-from util import lowercase, capitalize, normalize, Document, SchemaType
+from util import lowercase, capitalize, normalize, Document
 import os
 from protobuf_types import is_proto_field_scalar, get_proto_field_cpp_type, get_proto_field_cpp_prototype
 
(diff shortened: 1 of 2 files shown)


xml_converter/generators/generate_cpp.py
```diff
@@ -260,71 +269,73 @@ def generate_cpp_variable_data(
...
-                class_name = documentation_type_data[fieldval['type']]["class_name"]
+            if fieldval.variable_type in documentation_type_data:
+                cpp_type = documentation_type_data[fieldval.variable_type]["cpp_type"]
+                class_name = documentation_type_data[fieldval.variable_type]["class_name"]
                 cpp_includes.cpp_relative_includes.add("attribute/{}.hpp".format(class_name))
 
-            elif fieldval['type'] in ["Custom", "CompoundCustomClass"]:
-                cpp_type = fieldval['class']
-                class_name = insert_delimiter(fieldval['class'], delimiter="_")
+            # elif fieldval.variable_type in ("Custom", "CompoundCustomClass"):
```
klingbolt, line 286: This comment seems to be an alternative to 287. Should it be deleted?

xml_converter/generators/generate_cpp.py
```diff
@@ -260,71 +269,73 @@ def generate_cpp_variable_data(
...
+            elif fieldval.variable_type == "Custom" or fieldval.variable_type == "CompoundCustomClass":
+                cpp_type = fieldval.cpp_class
+                class_name = insert_delimiter(fieldval.cpp_class, delimiter="_")
                 cpp_includes.hpp_relative_includes.add("attribute/{}.hpp".format(class_name))
 
-            elif fieldval['type'] in ["Enum", "MultiflagValue", "CompoundValue"]:
+            elif fieldval.variable_type in ("Enum", "MultiflagValue", "CompoundValue"):
```
klingbolt, line 292: Is there a reason to use a Tuple here over a list?

xml_converter/generators/generate_cpp.py
```diff
@@ -260,71 +269,73 @@ def generate_cpp_variable_data(
...
             proto_drilldown_calls: str
             mutable_proto_drilldown_calls: str
             protobuf_field: str
-            proto_drilldown_calls, mutable_proto_drilldown_calls, protobuf_field = split_field_into_drilldown(fieldval["protobuf_field"])
+            proto_drilldown_calls, mutable_proto_drilldown_calls, protobuf_field = split_field_into_drilldown(fieldval.protobuf_field)
 
             # Compound Values are unique in that the components have xml fields in addition to the compound variable
-            if fieldval['type'] in ["CompoundValue", "CompoundCustomClass"]:
-                for component in fieldval['components']:
+            # if fieldval.variable_type in ("CompoundValue", "CompoundCustomClass"):
```
klingbolt, line 314: This comment seems to be an alternative to 315. Should it be deleted?

xml_converter/generators/generate_cpp.py
AsherGlick: I left this here because the code in 287 works for type inference, but for some reason the code in 286 does not. They should be equal and I like the style of 286 better.

This might not be an issue in mypy but it is in the LSP I am using for my editor pyright. I think both the comment and the "working" line should stay.

xml_converter/generators/generate_cpp.py
AsherGlick: Conceptually no. However tuples are immutable while lists are mutable. So in theory python could do some sort of optimization. No reason to force tuples over lists but they should probably be preferred.

klingbolt (approved):

PR #303, opened 2024-05-26
Adding the ability to add examples to component fields

AsherGlick:
Also fixes the XML attribute named used in the examples

### DIFF
--- a/xml_converter/doc/position/position.md
+++ b/xml_converter/doc/position/position.md
@@ -17,7 +17,7 @@ components:
   type: Float32
   xml_fields: [YPos, PositionY]
   protobuf_field: "y"
-  examples: ["81.727", "24.5189", "9.12564", "3.91473"]
+  examples: ["81.727", "24.5189", "-9.12564", "3.91473"]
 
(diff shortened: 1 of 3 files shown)


xml_converter/doc/position/position.md
```diff
@@ -11,16 +11,20 @@ components:
   type: Float32
   xml_fields: [XPos, PositionX]
   protobuf_field: "x"
+  examples: ["166.735", "1110.4", "-433.615", "-196.412"]
 
 - name: Y Position
   type: Float32
   xml_fields: [YPos, PositionY]
   protobuf_field: "y"
+  examples: ["81.727", "24.5189", "9.12564", "3.91473"]
```
klingbolt, line 20: One of these should be negative since they are valid values.

klingbolt (changes requested):

xml_converter/doc/position/position.md
AsherGlick: Done

klingbolt (approved):

PR #318, opened 2024-06-17
Changed all instances of the word path or route to Trail

klingbolt:
Route and path were used interchangeably in the gdscript code. This PR changes them all to Trail which is the naming used in the xml converter and protobuf schema.

PR #319, opened 2024-06-23
Changed Textures to be assigned by waypoint texture ID

klingbolt:
Also move textures into the user data directory and assign new ID. XML converter will handle only saving the images that are used by the icons and trails so single deletion is not necessary at this time.


Spatial.gd
```diff
@@ -533,21 +534,10 @@ func _waypoint_categories_to_godot_nodes(item: TreeItem, waypoint_category: Wayp
...
 
 	for trail in waypoint_category.get_trail():
-		var texture_id = trail.get_texture_id()
-		if texture_id == null:
-			print("Warning: No texture found in " , category_name)
-			continue
-		var full_texture_path = self.marker_file_dir + self.waypoint_data.get_textures()[texture_id].get_filepath()
-		gen_new_trail(full_texture_path, trail, category_item)
-
+		gen_new_path(trail, category_item)
```
klingbolt, line 537:
```suggestion
		gen_new_trail(trail, category_item)
```

Spatial.gd
```diff
@@ -908,7 +909,12 @@ func _on_ChangeTexture_pressed():
 # or icon is created.
 ################################################################################
 func _on_TexturePathOpen_file_selected(path):
```
klingbolt, line 911:
```suggestion
func _on_TexturePathOpen_file_selected(path: String):
```

Spatial.gd
```diff
@@ -929,7 +935,14 @@ func _on_NewIcon_pressed():
 	position.set_x(self.player_position.x)
 	position.set_y(self.player_position.y)
 	position.set_z(-self.player_position.z)
-	gen_new_icon(self.next_texture_trail, waypoint_icon, self.currently_active_category)
+	var texture_id: int = get_texture_index(self.next_texture_path)
```
AsherGlick, line 938: Why are we doing this texture index lookup every time we add a new icon, instead of having a cached`self.next_texture_index` variable like we have with next_texture_path

Spatial.gd
```diff
@@ -929,7 +935,14 @@ func _on_NewIcon_pressed():
 	position.set_x(self.player_position.x)
 	position.set_y(self.player_position.y)
 	position.set_z(-self.player_position.z)
-	gen_new_icon(self.next_texture_trail, waypoint_icon, self.currently_active_category)
```
AsherGlick, line 932: wat
https://github.com/AsherGlick/Burrito/pull/318/files#diff-de757fecd7208a343c7d1d1ac26b99a2a87d2f8f61537b01a868f0e058441bbcR932
How did this pass testing in the last PR?

Spatial.gd
```diff
@@ -929,7 +935,14 @@ func _on_NewIcon_pressed():
...
+	var texture_id: int = get_texture_index(self.next_texture_path)
+	if texture_id != -1:
+		waypoint_icon.set_texture_id(texture_id)
+	else:
+		self.waypoint_data.get_textures().add_textures().set_filepath(self.next_texture_path)
+		texture_id = self.waypoint_data.get_textures().add_textures().size() - 1
+		waypoint_icon.set_texture_id(texture_id)
```
AsherGlick, line 944: This wont be necessary here if we already have a `self.next_texture_index`

Spatial.gd
```diff
@@ -943,7 +956,14 @@ func _on_NewTrailPoint_pressed():
...
 		trail_data.add_points_z(-self.player_position.z)
-		var new_trails = gen_new_trail(self.next_texture_path, waypoint_trail, self.currently_active_category)
+		var texture_id: int = get_texture_index(self.next_texture_path)
+		if texture_id != -1:
+			waypoint_trail.set_texture_id(texture_id)
+		else:
+			self.waypoint_data.get_textures().add_textures().set_filepath(self.next_texture_path)
+			texture_id = self.waypoint_data.get_textures().add_textures().size() - 1
+			waypoint_trail.set_texture_id(texture_id)
+		var new_trails: PoolStringArray = gen_new_trail(waypoint_trail, self.currently_active_category)
```
AsherGlick, line 966: ditto the next_texture_index comments from on new icon pressed

FileHandler.gd
```diff
@@ -16,3 +16,61 @@ static func create_directory_if_missing(path: String):
...
+static func generate_unique_file_name(base_name: String, extension: String, destination_folder: String) -> String:
+	var unique_name: String = base_name
+	var counter: int = 1
+	var new_file_path: String = destination_folder.plus_file(unique_name + "." + extension)
+
+	while File.new().file_exists(new_file_path):
+		unique_name = base_name + "_" + str(counter)
+		new_file_path = destination_folder.plus_file(unique_name + "." + extension)
+		counter += 1
+	return unique_name + "." + extension
```
AsherGlick, line 29: This function is fine, but it deserves a comment with some input/output examples. This is a prime candidate for some unit tests once we get around to adding those to gdscript code because it is a simple function with strict inputs and outputs.

FileHandler.gd
```diff
@@ -16,3 +16,61 @@ static func create_directory_if_missing(path: String):
...
+
+static func copy_supported_image_file(file_path: String, destintation_dir: String) -> String:
+	# These are all supported file types in Godot but may not be supported
+	# by other marker programs.
+	var supported_image_extensions: Array = ["png", "jpg", "jpeg", "bmp", "tga"]
```
AsherGlick, line 34: We should check what other overlay tools support. 
My quick scan of known marker packs has these stats:
.png - ~12k files
.jpeg - 0 files
.jpg - 1 file named "jpgguide.png"
.bmp - 1
.tga - several "mythwrigh**tga**mbit...png"

FileHandler.gd
```diff
@@ -16,3 +16,61 @@ static func create_directory_if_missing(path: String):
...
+		var destination_mod_time: int = file.get_modified_time(file_name)
+		file.close()
+
+		if original_size == destination_size and original_mod_time == destination_mod_time:
+			return destination_path
+		else:
+			var base_name: String = file_name.get_basename()
+			var extension: String = file_name.get_extension()
+			var unique_name: String = generate_unique_file_name(base_name, extension, destintation_dir)
+			destination_path = destintation_dir.plus_file(unique_name)
```
AsherGlick, line 62: This needs to be thought through some more.

FileHandler.gd
```diff
@@ -16,3 +16,61 @@ static func create_directory_if_missing(path: String):
...
+
+	var result: int = file.open(destination_path, File.WRITE)
+	if result == OK:
+		file.store_buffer(content)
+		file.close()
+		print("File imported successfully.")
+		return destination_path
+	else:
+		print("Failed to import file.")
+		return ""
```
AsherGlick, line 76: This needed some more research before doing it by rote.
https://docs.godotengine.org/en/3.2/classes/class_directory.html?file#class-directory-method-copy

AsherGlick (changes requested):
Overall your PRs seem to be improving the last couple weeks. Though this one should have been 2 PRs. One of those PRs, reading the texture path from the waypoint data, only needed some very minor changes. The other one, needs some more design work and user workflow consideration before merging.

FileHandler.gd
klingbolt: Does it make sense to move the part where it checks if the file is the same to a different function and leave this one as just copying? My thought process is that if a user tries to selected the same image twice, they wouldn't care to know that it was already in the user data. But that is a functional determination and doesn't need to be part of this function. 

Or is the issue in how I determined if the two files are identical?

FileHandler.gd
klingbolt: Will add a comment

Spatial.gd
klingbolt: Not sure but it's being deleted.

Spatial.gd
klingbolt: Switching to `self.next_texture_index` variable. This variable will need to be updated on map changes but testing should that for 100 textures, the time to search was 180 microseconds.

Spatial.gd
klingbolt: Agreed

FileHandler.gd
klingbolt: After consideration, I believe it is better design to have the user save the new file. We can default to the old name for the file but this will let the user organize the file if they wish.

Spatial.gd
```diff
@@ -343,11 +343,19 @@ func decode_context_packet(spb: StreamPeerBuffer):
 
 	if self.map_id != old_map_id:
 		print("New Map")
+		var old_textue_path: String = ""
 		if old_map_id != 0 and not read_hash(old_map_id) == make_hash(self.waypoint_data.to_bytes()):
 			print("Saving Old Map")
+      old_textue_path = self.waypoint_data.get_textures()[self.next_texture_id]
```
AsherGlick, line 349: tabs vs spaces

Spatial.gd
```diff
@@ -343,11 +343,19 @@ func decode_context_packet(spb: StreamPeerBuffer):
 
 	if self.map_id != old_map_id:
 		print("New Map")
+		var old_textue_path: String = ""
 		if old_map_id != 0 and not read_hash(old_map_id) == make_hash(self.waypoint_data.to_bytes()):
 			print("Saving Old Map")
+			old_textue_path = self.waypoint_data.get_textures()[self.next_texture_id].get_filepath()
```
AsherGlick, line 349: Why are we doing this inside this if statement?

Spatial.gd
```diff
@@ -343,11 +343,19 @@ func decode_context_packet(spb: StreamPeerBuffer):
...
 			save_map_data(old_map_id)
 		print("Loading New Map")
 		load_waypoint_markers(self.map_id)
+		if old_textue_path != "":
+			self.next_texture_id = get_texture_index(old_textue_path)
+			if self.next_texture_id == -1:
+				self.waypoint_data.add_textures().set_filepath(old_textue_path)
+				self.next_texture_id = self.waypoint_data.get_textures().size() - 1
```
AsherGlick, line 357: I think there should be an else here to reset or unset next_texture_id

Spatial.gd
```diff
@@ -577,14 +575,18 @@ func apply_category_visibility_to_nodes(category_item: TreeItem):
...
+func gen_new_trail(waypoint_trail: Waypoint.Trail, category_item: TreeItem) -> Array:
 	# Create the texture to use from an image file
 	# TODO: We want to be able to cache this data so that if a texture is used
 	# by multiple objects we only need to keep ony copy of it in memory. #22.
 	# TODO: We want to have two copies of each texture in memory one for 2D
 	# which does not use srgb to render properly, and one for 3D which forces
 	# srgb to render properly. Issue #23.
-	# TODO: Search within waypoint data for texture paths instead of input
+	var texture_id: int = waypoint_trail.get_texture_id()
+	if texture_id == null:
```
AsherGlick, line 586: Offline testing we did says that this can never be null

Spatial.gd
```diff
@@ -614,13 +615,18 @@ func gen_new_trail(texture_path: String, waypoint_trail: Waypoint.Trail, categor
 	new_trail2d.refresh_points()
 	category_data.category2d.add_trail2d(new_trail2d)
 
-	return [new_trail3d, new_trail2d]
+	return Array([new_trail3d, new_trail2d])
```
AsherGlick, line 618: Why are we wrapping an array with Array

Spatial.gd
```diff
@@ -614,13 +615,18 @@ func gen_new_trail(texture_path: String, waypoint_trail: Waypoint.Trail, categor
...
 
-func gen_new_icon(texture_path: String, waypoint_icon: Waypoint.Icon, category_item: TreeItem):
-	# TODO: Search within waypoint data for texture paths instead of input
+func gen_new_icon(waypoint_icon: Waypoint.Icon, category_item: TreeItem):
+	var texture_id: int = waypoint_icon.get_texture_id()
+	if texture_id == null:
```
AsherGlick, line 622: 0

Spatial.gd
```diff
@@ -930,8 +944,23 @@ func _on_ChangeTexture_pressed():
 # Set the file that will be used to create a new trail or icon when a new trail
 # or icon is created.
 ################################################################################
-func _on_TexturePathOpen_file_selected(path):
-	self.next_texture_path = path
+var temp_image_file_path: String = ""
```
AsherGlick, line 947: Is this used anywhere

FileHandler.gd
```diff
@@ -16,3 +16,51 @@ static func create_directory_if_missing(path: String):
...
+			if file_name == target_name:
+				var file = File.new()
+				file.open(full_path, File.READ)
+				var file_content = file.get_buffer(file.get_len())
+				file.close()
+				if file_content == target_content:
+					dir.list_dir_end()
+					return current_relative_path
+			file_name = dir.get_next()
+		return ""
```
AsherGlick, line 42: return null from this function when a file is not found instead of an empty string.

FileHandler.gd
```diff
@@ -16,3 +16,51 @@ static func create_directory_if_missing(path: String):
...
+
+static func find_image_duplicates(file_path: String, destintation_dir: String) -> String:
+	# These are all supported file types in Godot that are not supported
+	# by other marker programs. "jpg", "jpeg", "bmp", "tga"
```
AsherGlick, line 46: We don't need this comment explaining what formats godot supports.

FileHandler.gd
```diff
@@ -16,3 +16,51 @@ static func create_directory_if_missing(path: String):
 		var success = dir.make_dir(path)
 		if success != OK:
 			print("Error: Could not create data folder:", path)
+
+static func find_file_duplicate(directory_path: String, target_name: String, target_content: PoolByteArray, relative_path: String):
+	var dir = Directory.new()
+	if dir.open(directory_path) == OK:
```
AsherGlick, line 22: As discussed offline, switch this logic so you can unindent the code below.

AsherGlick (changes requested):

AsherGlick (approved):

PR #321, opened 2024-07-02
Added a test for missing texture files

klingbolt:
This test does not have files to support the texture paths.

Double errors are expected as there is a check when writing the proto and a check when writing the xml


AsherGlick:
What is the purpose of this test? I dont see any associated bug or fix related to it.

klingbolt:
> What is the purpose of this test? I dont see any associated bug or fix related to it.

The bug I was trying to was eventually solved by #320
I had two purposes for this. 
- I wanted a file with a large number of textures so that i could evaluate performance while working on the #319. 
- I wanted to ensure the expected behavior which is that the path in the texture field that does not refer to a real file is still passed in and doesn't leave an empty texture array.

AsherGlick:
Unit tests or integration tests are not generally a good place to put performance tests. Performance tests are by definition slow and will just end up slowing down the unit tests.

Ensuring that the expected behavior occurs as described can be done with a smaller better named test then the one written in this PR. Effectively this test is just checking the value of IconFile 100 times. Which would be the same test if it was done a single time instead.

klingbolt:
Makes sense. I'll change this to a smaller test then

xml_converter/integration_tests/test_cases/texture_invalid/testcase.yaml
```diff
@@ -0,0 +1,7 @@
+input_paths: 
+    "pack": "xml"
+expected_stdout: |
+    Warning: File path test_cases/texture_invalid/input/pack/texture_01.png not found.
+    Warning: File path test_cases/texture_invalid/input/pack/texture_01.png not found.
```
AsherGlick, line 5: Why is this printed twice?

klingbolt: Double errors are expected as there is a check when writing the proto and a check when writing the xml

PR #323, opened 2024-07-24
Save the current data to the map files

klingbolt:
This covers the third issue in #267.

Saving the waypoint data to file

### DIFF
--- a/Spatial.gd
+++ b/Spatial.gd
@@ -9,6 +9,7 @@ var map_is_open: bool
 var compass_is_top_right: bool
 
 var edit_panel_open: bool = false
+var unsaved_changes: bool = false setget set_unsaved_changes
 
 # This is the path to the texture that will be used for the next created 3d-path
 # object or icon object in the UI
@@ -389,11 +390,14 @@ func reset_3D_minimap_masks(category: Spatial):
 
 
 var waypoint_data = Waypoint.Waypoint.new()
-var marker_file_dir = "user://protobins/"
+#We save the marker data in this directory when the files are have been split
+#by Map ID. All changes made by the editor are automatically saved in these
+#files prior to export.
+var unsaved_markers_dir = "user://protobins/"
 var marker_file_path = ""
 
-func load_waypoint_markers(map_id):
-	self.marker_file_path = self.marker_file_dir + String(map_id) + ".bin"
+func load_waypoint_markers(map_id_to_load: int):
+	self.marker_file_path = self.unsaved_markers_dir + String(map_id_to_load) + ".bin"
 	self.waypoint_data = Waypoint.Waypoint.new()
 	clear_map_markers()
 	init_category_tree()
@@ -537,7 +541,7 @@ func _waypoint_categories_to_godot_nodes(item: TreeItem, waypoint_category: Wayp
 		if texture_id == null:
 			print("Warning: No texture found in " , category_name)
 			continue
-		var full_texture_path = self.marker_file_dir + self.waypoint_data.get_textures()[texture_id].get_filepath()
+		var full_texture_path = self.unsaved_markers_dir + self.waypoint_data.get_textures()[texture_id].get_filepath()
 		gen_new_trail(full_texture_path, trail, category_item)
 
 
@@ -546,7 +550,7 @@ func _waypoint_categories_to_godot_nodes(item: TreeItem, waypoint_category: Wayp
 		if texture_id == null:
 			print("Warning: No texture found in " , category_name)
 			continue
-		var full_texture_path = self.marker_file_dir + self.waypoint_data.get_textures()[texture_id].get_filepath()
+		var full_texture_path = self.unsaved_markers_dir + self.waypoint_data.get_textures()[texture_id].get_filepath()
 		gen_new_icon(full_texture_path, icon, category_item)
 
 	for category_child in waypoint_category.get_children():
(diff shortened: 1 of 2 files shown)


Spatial.gd
```diff
@@ -849,14 +855,19 @@ func new_trail_point_after(waypoint_trail: Waypoint.Trail, trail3d: Spatial, tra
 
 func refresh_trail3d_points(trail3d: Spatial):
 	trail3d.refresh_mesh()
+	set_unsaved_changes(true)
```
AsherGlick, line 858: This seems like the wrong place to trigger the fact that the proto data has been changed.

Spatial.gd
```diff
@@ -849,14 +855,19 @@ func new_trail_point_after(waypoint_trail: Waypoint.Trail, trail3d: Spatial, tra
...
 
 func refresh_trail2d_points(trail2d: Line2D):
 	trail2d.refresh_points()
+	set_unsaved_changes(true)
```
AsherGlick, line 862: This seems like the wrong place to trigger the fact that the proto data has been changed.

Spatial.gd
```diff
@@ -849,14 +855,19 @@ func new_trail_point_after(waypoint_trail: Waypoint.Trail, trail3d: Spatial, tra
...
 
 
 ################################################################################
 # Signal Functions
 ################################################################################
+func _on_SaveTrail_pressed():
+	save_current_map_data()
+
```
AsherGlick, line 870: Isn't this button supposed to take data from the split files and save it to the combined files, or in the current temporary case: "the one big combined file which we will eventually split out into individual combined files"

Spatial.gd
```diff
@@ -1011,6 +1022,8 @@ func _on_ReverseTrailDirection_pressed():
 
 
 func _on_ExitButton_pressed():
+	if self.unsaved_changes:
+		save_current_map_data()
```
AsherGlick, line 1026: What happens if a current or future bug fails to call `set_unsaved_changes(true)`

Spatial.gd
```diff
@@ -389,11 +392,14 @@ func reset_3D_minimap_masks(category: Spatial):
 
 
 var waypoint_data = Waypoint.Waypoint.new()
-var marker_file_dir = "user://protobins/"
+#We save the marker data in this directory when the files are have been split
+#by Map ID. All changes made by the editor are automatically saved in these
+#files prior to export.
```
AsherGlick, line 397: Nit: Put a space after each `#`

Spatial.gd
```diff
@@ -9,6 +9,7 @@ var map_is_open: bool
 var compass_is_top_right: bool
 
 var edit_panel_open: bool = false
+var unsaved_changes: bool = false setget set_unsaved_changes
```
AsherGlick, line 12: Is this a `set` or a `setget` function?

AsherGlick (changes requested):
I had to search through the linked issue to find what the "third issue" was. I am assuming it is from this comment: https://github.com/AsherGlick/Burrito/pull/267#pullrequestreview-2013191899

Did we change how we designed saving to work or am totally missing something in this PR?

Spatial.gd
```diff
@@ -339,7 +340,11 @@ func decode_context_packet(spb: StreamPeerBuffer):
 
 	if self.map_id != old_map_id:
 		print("New Map")
-		print("Saving Old Map")
+		if old_map_id != 0:
+			if not read_hash(old_map_id) == get_hash(self.waypoint_data.to_bytes()):
```
AsherGlick, line 344: These two lines might be able to be combined with an `and` or `&&` I forget which syntax gdscript uses, most languages stop evaluating an `and` if the left half is false.

Spatial.gd
```diff
@@ -339,7 +340,11 @@ func decode_context_packet(spb: StreamPeerBuffer):
...
+				print("Saving Old Map")
+				save_map_data(old_map_id)
+				print("not same")
```
AsherGlick, line 347: We probably dont need this print

Spatial.gd
```diff
@@ -857,6 +883,9 @@ func refresh_trail2d_points(trail2d: Line2D):
 ################################################################################
 # Signal Functions
 ################################################################################
+func _on_SaveTrail_pressed():
+	save_map_data(self.map_id)
+
```
AsherGlick, line 888: Where is this called? It seems to be new to this PR but nothing in this PR calls it.

Spatial.gd
```diff
@@ -621,31 +629,49 @@ func gen_new_icon(texture_path: String, waypoint_icon: Waypoint.Icon, category_i
...
+	file.store_buffer(packed_bytes)
+
+func get_hash(data: PoolByteArray) -> String:
+	var ctx = HashingContext.new()
+	ctx.start(HashingContext.HASH_MD5)
+	ctx.update(data)
+	var res: PoolByteArray = ctx.finish()
+	return res.hex_encode()
+
+const FILE_HASH_PATH: String = "user://file_hash.json"
```
AsherGlick, line 648: Lets give this a more descriptive variable name and value, and move it to the top of the file.

Spatial.gd
```diff
@@ -621,31 +629,49 @@ func gen_new_icon(texture_path: String, waypoint_icon: Waypoint.Icon, category_i
...
+################################################################################
+# Section of functions for saving data to file
+################################################################################
+func save_map_data(map_id: int):
+	var packed_bytes = self.waypoint_data.to_bytes()
+	var file = File.new()
+	file.open(self.marker_file_path, file.WRITE)
+	file.store_buffer(packed_bytes)
+
+func get_hash(data: PoolByteArray) -> String:
```
AsherGlick, line 641: Just because the other to functions for save and load exist lets name this "make_hash" or something similar to indicate that it is not just being read. It would become obvious to anyone calling the function what it did but I think renaming it will make it a little clearer for people who are just reading the function call.

Spatial.gd
```diff
@@ -621,31 +629,49 @@ func gen_new_icon(texture_path: String, waypoint_icon: Waypoint.Icon, category_i
...
+# Save all hashes
+func save_hashes():
+	var file = File.new()
+	var data = {}
+	var dir = Directory.new()
+	dir.open(self.unsaved_markers_dir)
+	dir.list_dir_begin()
+	var file_name = dir.get_next()
+	while file_name != "":
+		if dir.file_exists(file_name):
```
AsherGlick, line 659: This seems like a weird check:
"for each file in the directory if the file exists do a thing"

Spatial.gd
```diff
@@ -621,31 +629,49 @@ func gen_new_icon(texture_path: String, waypoint_icon: Waypoint.Icon, category_i
...
+	file.store_string(JSON.print(data))
+
+func read_hash(map_id: int) -> String:
+	var file = File.new()
+	if not file.file_exists(FILE_HASH_PATH):
+		return ""
+	var data = {}
+	file.open(FILE_HASH_PATH, File.READ)
+	data = JSON.parse(file.get_as_text()).result
+	return data.get(String(map_id), "")
```
AsherGlick, line 674: You dont need to define data, then open the file, then set data. It could be 
```gdscript
	file.open(FILE_HASH_PATH, File.READ)
	var data = JSON.parse(file.get_as_text()).result
	return data.get(String(map_id), "")
```
or even
```gdscript
	file.open(FILE_HASH_PATH, File.READ)
	return JSON.parse(file.get_as_text()).result.get(String(map_id), "")
```
based on readability goals

AsherGlick (changes requested):

Spatial.gd
klingbolt: The while loop here also returns the name of subdirectories. The other option here was to use `if not dir.current_is_dir()` as there is not `current_is_file` or such.

Spatial.gd
klingbolt: I like that name change

Spatial.gd
klingbolt: Will do.

Spatial.gd
AsherGlick: I see, I think it would be better to use something like `is_file` unless `file_exists` is the only way to do that.

Spatial.gd
AsherGlick: I don't see it in the negative diff, I think it is being created in this PR. I think if it is not being used then it should be removed form this PR and added to whatever PR is actually going to use it.

klingbolt: Oh sorry I took it as this commit. Thats understandable, I've remove for now

Spatial.gd
klingbolt: There is no `is_file` function from what I can find.

AsherGlick: Then the way you have it is probably the right thing to do, the godot syntax is just weird. Not surprising given how weird the Directory/File syntax is.

PR #344, opened 2024-08-29
Single File for saving imported packs

klingbolt:
This changes the saved files to contain one single protobin file. This file is the definitive file and any changes to the internal data can be saved to it. 

Changed import function to add new pack into existing packs instead of overwriting.

A bug emerged where filesystem was faulting when both paths were identical. I thought this would have been covered by the copy_options.


AsherGlick:
As discussed offline, this code was not tested and does not function as intended. It should be corrected and tested before being sent for re-review.

AsherGlick (approved):

PR #347, opened 2024-09-05
Invert default visibility

klingbolt:
Changed Default_toggle to hide_category in the proto so the default can be 0. #151


klingbolt:
I am still uncertain if hide_category or something like hide_by_default. I felt like adding the word default is unnecessary but on the other hand, the current behavior of Burrito is to hide all categories by default so we may want to change this behavior in the future.

xml_converter/proto/waypoint.proto
```diff
@@ -21,7 +21,7 @@ message Category {
     repeated Icon icon = 3;
     repeated Trail trail = 4;
     bool is_separator = 5;
-    bool default_visibility = 6;
+    bool hide_category = 6;
```
AsherGlick, line 24: What were other names that were considered here and why did you decide on `hide_category`?
As it is written now this field is `Category.hide_category` which seems redundant.
Just removing `category` and making it `Category.hide` might look like a good idea initially but will it make intuitive sense when combined with the hidden/shown safe file data too?

klingbolt: The other top names I considered were `hide_by_default' and `default_hide_category`.
There is another field in the XML called `hide` that we made `trigger.action_hide_category` so I used that naming convention. Though on further thought, that field is a Category object and this one is a bool so maybe they don't need to be similar. 
I considered using `hide` but was concerned that it was too similar to the Godot function `hide()` and that when accessing this field, `Category.get_hide()` isn't descriptive enough.

AsherGlick: My gut tells me that `hide_category` is too imperative / verb-ish when it should be more declarative / noun-ish.

klingbolt: I agree

How about `is_hidden`? This makes it clear its a property not a command. Also fits with some other attributes like `is_separator` and `is_wall`.

PR #353, opened 2024-09-11
Fixed a bug that caused trails to not inherit attributes correctly

klingbolt:


xml_converter/src/packaging_xml.cpp
```diff
@@ -168,7 +168,7 @@ vector<Parseable*> parse_pois(rapidxml::xml_node<>* root_node, map<string, Categ
             for (size_t category_index = 0; category_index < categories.size(); category_index++) {
                 for (size_t i = 0; i < categories[category_index]->icon_attributes.size(); i++) {
```
AsherGlick, line 169: looks like more robust tests are needed here

AsherGlick (changes requested):

xml_converter/src/packaging_xml.cpp
klingbolt: Good catch. The test seems to have only worked because the size of the two arrays was the same.

xml_converter/integration_tests/test_cases/xml_category_inheritance/input/pack/xml_file.xml
```diff
@@ -1,7 +1,7 @@
 <OverlayData>
     <MarkerCategory Name="mycategory" XPos="169.81" YPos="210.65" ZPos="215.83"></MarkerCategory>
 
-    <MarkerCategory Name="NestedLevel1" XPos="000" MapID="999">
+    <MarkerCategory Name="NestedLevel1" XPos="000" MapID="999" IconFile="texture_one.png" Texture="texture_one.png" TrailData="trail.trl" AnimationSpeed="1">
```
AsherGlick, line 4: What do you think about spiting the category into two instead of adding arbitrary arguments that correspond to trails vs icons. This way we wont accidentally run into this issue in the future if new required attributes are added to the schema.

To me it seems like this solution as written is a "if we add these irrelevant fields it happens to test the thing we want".

AsherGlick (changes requested):

xml_converter/integration_tests/test_cases/xml_category_inheritance/input/pack/xml_file.xml
klingbolt: I hadn't considered the required attributes. 

My thought in adding them to this category was to ensure that the attributes were sorted correctly into either Icons or Trails.

AsherGlick (approved):

PR #357, opened 2024-09-23
Added tracking for the top level categories in each pack. 

klingbolt:
Stops writing if there are any duplicates between packs. Added an argument to bypass this if necessary. Added tests for both cases


xml_converter/src/xml_converter.cpp
```diff
@@ -144,20 +152,29 @@ void process_data(
...
     for (size_t i = 0; i < input_taco_paths.size(); i++) {
         cout << "Loading taco pack " << input_taco_paths[i] << endl;
 
-        read_taco_directory(
+        set<string> category_names = read_taco_directory(
             input_taco_paths[i],
             &marker_categories,
             &parsed_pois);
+        for (string category_name : category_names) {
+            if (find(top_level_categories.begin(), top_level_categories.end(), category_name) != top_level_categories.end())
```
AsherGlick, line 173: Is this the right function? I would think it would be `top_level_categories.find(category_name)`. Where is this `find()` function from and why do you need to pass `begin()` and `end()` iterators into it?

xml_converter/src/xml_converter.cpp
```diff
@@ -144,20 +152,29 @@ void process_data(
...
     auto begin = chrono::high_resolution_clock::now();
     for (size_t i = 0; i < input_taco_paths.size(); i++) {
         cout << "Loading taco pack " << input_taco_paths[i] << endl;
 
-        read_taco_directory(
+        set<string> category_names = read_taco_directory(
             input_taco_paths[i],
             &marker_categories,
             &parsed_pois);
+        for (string category_name : category_names) {
```
AsherGlick, line 172: The tooling to check if a category is a duplicate or not should probably be broken out into its own function.

xml_converter/src/xml_converter.cpp
```diff
@@ -168,10 +185,27 @@ void process_data(
     for (size_t i = 0; i < input_guildpoint_paths.size(); i++) {
         cout << "Loading guildpoint pack " << input_guildpoint_paths[i] << endl;
 
-        read_burrito_directory(
+        set<string> category_names = read_burrito_directory(
             input_guildpoint_paths[i],
             &marker_categories,
             &parsed_pois);
+        for (string category_name : category_names) {
+            if (find(top_level_categories.begin(), top_level_categories.end(), category_name) != top_level_categories.end())
```
AsherGlick, line 193: Ditto on the earlier `find()` question

xml_converter/src/xml_converter.cpp
```diff
@@ -168,10 +185,27 @@ void process_data(
...
+            else
+                top_level_categories.insert(category_name);
+        }
+    }
+
+    if (duplicate_categories.size() > 0) {
+        cout << "The following top level categories were found in more than one pack" << endl;
+        for (size_t i = 0; i < duplicate_categories.size(); i++) {
+            cout << duplicate_categories[i] << endl;
+        }
```
AsherGlick, line 204: Do we want to print out all the duplicates if we are allowing for duplicates? Also just checking top-level categories means that we will only detect top level category duplicates. Which is enough to answer "are there duplicates" but not "what duplicates are there".

xml_converter/src/xml_converter.cpp
```diff
@@ -144,20 +152,29 @@ void process_data(
 
     // This is a special output path used for burrito internal use that splits
     // the guildpoint protobins by map id.
-    string output_split_guildpoint_dir) {
+    string output_split_guildpoint_dir,
+    bool allow_duplicates) {
```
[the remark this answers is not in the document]
klingbolt: It is. I'll move it.

xml_converter/src/xml_converter.cpp
klingbolt: That's a good point. I will move the check. 

I think we only want to check top level duplicates. If there are any duplicates in the children than the top level will be duplicated as well. If for instance we are running a pack twice, the list of duplicates would be every single category in the pack.  I.e there is no instance where `a.b.c` is duplicated where `a.b` and `a` are as well.

xml_converter/src/xml_converter.cpp
```diff
@@ -244,6 +279,9 @@ int main(int argc, char* argv[]) {
             // CLI arg parsing later to properly capture this.
             arg_target = &output_split_guildpoint_paths;
         }
+        else if (!strcmp(argv[i], "--allow-duplicates")) {
+            allow_duplicates = true;
+        }
```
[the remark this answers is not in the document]
klingbolt: That makes sense. I'll look at other programs and see what makes sense.

xml_converter/src/packaging_xml.cpp
```diff
@@ -82,9 +83,11 @@ void parse_marker_categories(
         for (rapidxml::xml_node<>* child_node = node->first_node(); child_node; child_node = child_node->next_sibling()) {
             parse_marker_categories(child_node, &(category->children), category, errors, state, depth + 1);
         }
+        return name;
     }
     else {
         errors->push_back(new XMLNodeNameError("Unknown MarkerCategory Tag", node));
+        return "";
```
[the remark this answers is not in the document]
klingbolt: I agree and I'll make this change

xml_converter/src/xml_converter.cpp
klingbolt: I agree. I'll spin out this function. I imagine this function would take in a pointer to the set and vector. Is this specific enough it should be in this file or should I include it in the `string_helper.hpp` functions

xml_converter/integration_tests/run_tests.py
```diff
@@ -20,6 +20,7 @@ def run_xml_converter(
     input_proto: Optional[List[str]] = None,
     output_proto: Optional[List[str]] = None,
     split_output_proto: Optional[str] = None,
+    additional_arguments: Optional[List[str]] = None,
```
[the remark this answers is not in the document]
klingbolt: I didn't catch the issue with it adding previous arguments. I will change this to explicitly for `allow-duplicates`

xml_converter/src/xml_converter.cpp
klingbolt: I made a more generic function that combines two sets and reports the duplicates between them.

xml_converter/src/packaging_xml.cpp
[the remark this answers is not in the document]
klingbolt: Added an OptionalString struct

xml_converter/src/xml_converter.cpp
AsherGlick: I'm guessing algorhtim's find is an o(n) linear search, which is much worse than set.finds(log(n)). Iirc set uses a tree not a hash table behind the scenes.

xml_converter/src/xml_converter.cpp
AsherGlick: as discussed offline: lets add some extra instructions to the user as to what to do when there are duplicates.

xml_converter/src/packaging_xml.cpp
```diff
@@ -17,49 +17,54 @@ using namespace std;
...
+        false,  // error
+    };
     if (get_node_name(node) == "MarkerCategory") {
-        string name;
-
         rapidxml::xml_attribute<>* name_attribute = find_attribute(node, "name");
         if (name_attribute == 0) {
             // TODO: This error should really be for the entire node not just the name
             errors->push_back(new XMLNodeNameError("Category attribute 'name' is missing so it cannot be properly referenced", node));
+            name.error = true;
```
AsherGlick, line 36: As discussed offline: This is not actually a "null" value or an error in the function processing.

xml_converter/src/packaging_xml.cpp
```diff
@@ -17,49 +17,54 @@ using namespace std;
...
 
             // TODO: Maybe fall back on display name slugification.
-            name = "UNKNOWN_CATEGORY_" + to_string(UNKNOWN_CATEGORY_COUNTER);
+            name.value = "UNKNOWN_CATEGORY_" + to_string(UNKNOWN_CATEGORY_COUNTER);
```
AsherGlick, line 39: I think that we should be setting the entire optional instead of just parts of it so it is easier to see which values are null vs not null.

xml_converter/src/string_helper.hpp
```diff
@@ -2,9 +2,15 @@
 
 #include <cstdint>
 #include <initializer_list>
+#include <set>
 #include <string>
 #include <vector>
 
+struct OptionalString {
+    std::string value;
+    bool error;
```
AsherGlick, line 11: I think `error` is an incorrect name to indicate if a value is null or not.

xml_converter/src/string_helper.cpp
```diff
@@ -314,3 +315,16 @@ std::string long_to_hex_string(uint64_t number) {
...
+    std::set<std::string>* set_b,
+    std::vector<std::string>* duplicates) {
+    for (string str : *set_a) {
+        if (auto search = set_b->find(str); search != set_b->end()) {
+            duplicates->push_back(str);
+        }
+        else
+            set_b->insert(str);
+    }
+}
```
AsherGlick, line 330: I don't think this function belongs in string_helper. It looks specific to the category name de-duplication.


If the name of the function is descriptive of what it is supposed to do, I think you can do:
```c++
set_a.insert(set_b.begin(), set_b.end())
```

If you are trying to find the intersection of the two sets instead you could use: https://en.cppreference.com/w/cpp/algorithm/set_intersection it seems.

Otherwise maybe this function should get a different name and live closer to where it is used.

xml_converter/src/xml_converter.cpp
```diff
@@ -189,23 +185,16 @@ void process_data(
...
-            if (find(top_level_categories.begin(), top_level_categories.end(), category_name) != top_level_categories.end())
-                duplicate_categories.push_back(category_name);
-            else
-                top_level_categories.insert(category_name);
-        }
+        combine_sets(&category_names, &top_level_categories, &duplicate_categories);
     }
 
-    if (duplicate_categories.size() > 0) {
+    if (duplicate_categories.size() > 0 && allow_duplicates != true) {
```
AsherGlick, line 191: `allow_duplicates != true`
should be
`allow_duplicates == false`

xml_converter/src/xml_converter.cpp
```diff
@@ -281,9 +270,16 @@ int main(int argc, char* argv[]) {
         }
         else if (!strcmp(argv[i], "--allow-duplicates")) {
             allow_duplicates = true;
+            arg_target = nullptr;
```
AsherGlick, line 273: We should probably change the default arg_target from input taco path to nullptr now that we have this logic.

AsherGlick (changes requested):

xml_converter/src/xml_converter.cpp
```diff
@@ -62,48 +62,62 @@ vector<string> get_files_by_suffix(string directory, string suffix) {
...
     if (!filesystem::exists(input_path)) {
         cout << "Error: " << input_path << " is not an existing directory or file" << endl;
     }
     else if (filesystem::is_directory(input_path)) {
+        string directory_name = filesystem::path(input_path).filename();
         vector<string> xml_files = get_files_by_suffix(input_path, ".xml");
         for (const string& path : xml_files) {
             set<string> category_names = (parse_xml_file(path, input_path, marker_categories, parsed_pois));
-            top_level_categories.insert(category_names.begin(), category_names.end());
+            string relative_path = directory_name + path.substr(input_path.size());
```
AsherGlick, line 78: Can we guarantee that we are actually getting the correct relative path here and not just stripping unrelated characters if for some reason `get_files_by_suffix()` changes it's output.

xml_converter/src/xml_converter.cpp
```diff
@@ -62,48 +62,62 @@ vector<string> get_files_by_suffix(string directory, string suffix) {
     return files;
 }
 
-set<string> read_taco_directory(
+map<string, vector<string>> read_taco_directory(
     string input_path,
     map<string, Category>* marker_categories,
     vector<Parseable*>* parsed_pois) {
-    set<string> top_level_categories;
+    map<string, vector<string>> category_to_filepath;
```
AsherGlick, line 69: Can we add a more descriptive name here. This is only storing the top level paths and mapping to the locations that they are saved in. Maybe `top_level_categories_to_file_locations` or something like that. Maybe just `top_level_category_file_locations` even.

xml_converter/src/xml_converter.cpp
```diff
@@ -62,40 +62,62 @@ vector<string> get_files_by_suffix(string directory, string suffix) {
...
     if (!filesystem::exists(input_path)) {
         cout << "Error: " << input_path << " is not an existing directory or file" << endl;
     }
     else if (filesystem::is_directory(input_path)) {
+        string directory_name = filesystem::path(input_path).filename();
         vector<string> xml_files = get_files_by_suffix(input_path, ".xml");
         for (const string& path : xml_files) {
-            parse_xml_file(path, input_path, marker_categories, parsed_pois);
+            set<string> category_names = (parse_xml_file(path, input_path, marker_categories, parsed_pois));
```
AsherGlick, line 77:
```suggestion
            set<string> top_level_category_names = parse_xml_file(path, input_path, marker_categories, parsed_pois);
```
nit: to be more specific on what category names we are getting. Also removing the extra parens.

xml_converter/src/xml_converter.cpp
```diff
@@ -62,40 +62,62 @@ vector<string> get_files_by_suffix(string directory, string suffix) {
...
+            string relative_path = directory_name + path.substr(input_path.size());
+            for (set<string>::iterator it = category_names.begin(); it != category_names.end(); it++) {
+                category_to_filepath[*it].push_back(relative_path);
+            }
         }
     }
     else if (filesystem::is_regular_file(input_path)) {
-        parse_xml_file(input_path, get_base_dir(input_path), marker_categories, parsed_pois);
+        set<string> category_names = parse_xml_file(input_path, get_base_dir(input_path), marker_categories, parsed_pois);
```
AsherGlick, line 85:
```suggestion
        set<string> top_level_category_names = parse_xml_file(input_path, get_base_dir(input_path), marker_categories, parsed_pois);
```
same nit from the directory branch

xml_converter/src/xml_converter.cpp
```diff
@@ -62,48 +62,62 @@ vector<string> get_files_by_suffix(string directory, string suffix) {
...
+    return category_to_filepath;
 }
 
-set<string> read_burrito_directory(
+map<string, vector<string>> read_burrito_directory(
     string input_path,
     map<string, Category>* marker_categories,
     vector<Parseable*>* parsed_pois) {
-    set<string> top_level_categories;
+    map<string, vector<string>> category_to_filepath;
```
AsherGlick, line 98: Same rename nit from the taco function's variable.

xml_converter/src/xml_converter.cpp
```diff
@@ -62,48 +62,62 @@ vector<string> get_files_by_suffix(string directory, string suffix) {
...
     if (!filesystem::exists(input_path)) {
         cout << "Error: " << input_path << " is not an existing directory or file" << endl;
     }
     else if (filesystem::is_directory(input_path)) {
+        string directory_name = filesystem::path(input_path).filename();
         vector<string> burrito_files = get_files_by_suffix(input_path, ".bin");
         for (const string& path : burrito_files) {
             set<string> category_names = read_protobuf_file(path, input_path, marker_categories, parsed_pois);
-            top_level_categories.insert(category_names.begin(), category_names.end());
+            string relative_path = directory_name + path.substr(input_path.size());
```
AsherGlick, line 107: same relative path nit

xml_converter/src/xml_converter.cpp
```diff
@@ -62,48 +62,62 @@ vector<string> get_files_by_suffix(string directory, string suffix) {
...
-    set<string> top_level_categories;
+    map<string, vector<string>> category_to_filepath;
     if (!filesystem::exists(input_path)) {
         cout << "Error: " << input_path << " is not an existing directory or file" << endl;
     }
     else if (filesystem::is_directory(input_path)) {
+        string directory_name = filesystem::path(input_path).filename();
         vector<string> burrito_files = get_files_by_suffix(input_path, ".bin");
         for (const string& path : burrito_files) {
             set<string> category_names = read_protobuf_file(path, input_path, marker_categories, parsed_pois);
```
AsherGlick, line 106: Same `category_names` nit

xml_converter/src/xml_converter.cpp
```diff
@@ -62,48 +62,62 @@ vector<string> get_files_by_suffix(string directory, string suffix) {
...
-            top_level_categories.insert(category_names.begin(), category_names.end());
+            string relative_path = directory_name + path.substr(input_path.size());
+            for (set<string>::iterator it = category_names.begin(); it != category_names.end(); it++) {
+                category_to_filepath[*it].push_back(relative_path);
+            }
         }
     }
     else if (filesystem::is_regular_file(input_path)) {
         set<string> category_names = read_protobuf_file(input_path, get_base_dir(input_path), marker_categories, parsed_pois);
```
AsherGlick, line 114: Same `category_names` nit

xml_converter/src/xml_converter.cpp
```diff
@@ -158,19 +172,22 @@ void process_data(
     // All of the loaded pois and categories
     vector<Parseable*> parsed_pois;
     map<string, Category> marker_categories;
-    set<string> top_level_categories;
-    vector<string> duplicate_categories;
+    map<string, vector<vector<string>>> category_to_marker_pack_names;
```
AsherGlick, line 175: same nit for this variable in the taco and burrito read functions.

xml_converter/src/xml_converter.cpp
```diff
@@ -158,19 +172,22 @@ void process_data(
...
+    map<string, vector<string>> duplicate_categories;
 
     // Read in all the xml taco markerpacks
     auto begin = chrono::high_resolution_clock::now();
     for (size_t i = 0; i < input_taco_paths.size(); i++) {
         cout << "Loading taco pack " << input_taco_paths[i] << endl;
 
-        set<string> category_names = read_taco_directory(
+        // swap this to map <category, set<filenames>>)
```
AsherGlick, line 183: Is this something you wanted to do in this pr? If not why not?

xml_converter/src/xml_converter.cpp
```diff
@@ -181,18 +198,37 @@ void process_data(
     for (size_t i = 0; i < input_guildpoint_paths.size(); i++) {
         cout << "Loading guildpoint pack " << input_guildpoint_paths[i] << endl;
 
-        set<string> category_names = read_burrito_directory(
+        map<string, vector<string>> category_names = read_burrito_directory(
```
AsherGlick, line 201: Same `category_names` nit

xml_converter/src/xml_converter.cpp
```diff
@@ -158,19 +172,22 @@ void process_data(
     // All of the loaded pois and categories
     vector<Parseable*> parsed_pois;
     map<string, Category> marker_categories;
-    set<string> top_level_categories;
-    vector<string> duplicate_categories;
+    map<string, vector<vector<string>>> category_to_marker_pack_names;
+    map<string, vector<string>> duplicate_categories;
```
AsherGlick, line 176: Should this be a map<string, set<string>> so that a filename does not appear twice if a category is present twice for some reason?

AsherGlick (changes requested):

xml_converter/src/xml_converter.cpp
klingbolt: Good point. I'll add a check to ensure that the output from `get_files_by_suffix()` continues to give me the format I expect.

xml_converter/integration_tests/test_cases/proto_and_xml_input_no_duplicates/testcase.yaml
```diff
@@ -3,9 +3,14 @@ input_paths:
...
-    Did not write due to duplicates in categories. If you want to bypass this, use '--allow-duplicates'
+    Did not write due to duplicates in categories.
+    This commonly occurs when attempting to read the same pack multiple times or when separate packs coincidentally have the same name.
+    Please remove one of the packs or edit the name of the packs' top level category before running the program again.
+    If you want to bypass this stop, use '--allow-duplicates'.
+    The following top level categories were found in more than one pack:
+    mycategory in files
+    pack/xml_file.xml
+    pack3/xml_file.xml
+    pack2/markers.bin
```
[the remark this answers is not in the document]
klingbolt: Yes it is deterministic. It does all the xml followed by all the proto inputs in the order that they are added as inputs to the program.

xml_converter/src/xml_converter.cpp
klingbolt: Changing the function get_files_by_suffix() to return the relative paths from the directory so that we can ensure it is the correct string.

xml_converter/src/xml_converter.cpp
klingbolt: This was a working comment. It is now the way i wanted but missed removing the comment

xml_converter/integration_tests/test_cases/proto_and_xml_input_no_duplicates/testcase.yaml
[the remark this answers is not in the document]
klingbolt: Changing the output to a set makes the list sorted. So it will now be in ascending alphabetical order

AsherGlick (approved):

PR #361, opened 2024-10-02
Commented out the attributes that are not implemented in Burrito

klingbolt:
This PR comments out the unoptimized proto fields so they can be used later. These fields are not currently used within Burrito so they can be added as they are optimized. 

More details behind the reason for this can be found in #151


AsherGlick:
As discussed offline we will:
* Use a new field in the markdown files instead of using `protobuf_field: null`
* We are also going to remove all fields that we are not 100% sure we know how they work and should be stored in a proto or are not going to be implementing soon.

xml_converter/integration_tests/src/testcase_loader.py
```diff
@@ -106,6 +106,19 @@ def load_testcase(path: str) -> Optional[Testcase]:
         print(f"Invalid Test, expecting string value for 'expected_returncode' in {path}")
         return None
 
+    expected_output_xml_path: str = os.path.join(path, "output_xml")
+    expected_output_proto_path: str = os.path.join(path, "output_proto")
+
+    if "exclude_output" in data:
+        for pack_type in data["exclude_output"]:
+            if pack_type == "xml":
+                expected_output_xml_path = ""
```
AsherGlick, line 115: This should be a unique value like `None` instead of treating `""` empty string as some special case path variable.

xml_converter/integration_tests/src/testcase_loader.py
```diff
@@ -106,6 +106,19 @@ def load_testcase(path: str) -> Optional[Testcase]:
...
+            if pack_type == "proto":
+                expected_output_proto_path = ""
```
AsherGlick, line 117: dito the special case `""`

xml_converter/integration_tests/run_tests.py
```diff
@@ -166,17 +166,24 @@ def main() -> bool:
...
+            os.makedirs(proto_output_dir_path, exist_ok=True)
+            output_proto_paths = [proto_output_dir_path]
 
         rawstdout, rawstderr, returncode = run_xml_converter(
             input_xml=testcase.xml_input_paths,
             input_proto=testcase.proto_input_paths,
-            output_xml=[xml_output_dir_path],
-            output_proto=[proto_output_dir_path],
+            output_xml=output_xml_paths if output_xml_paths else None,
+            output_proto=output_proto_paths if output_proto_paths else None,
```
AsherGlick, line 186: What is the difference here between passing in an empty list and passing in `None`? These lines seem to want to special case an `[]` empty list as the same thing as `None`.

xml_converter/integration_tests/run_tests.py
```diff
@@ -166,17 +166,24 @@ def main() -> bool:
...
                 continue
 
-        xml_output_dir_path = os.path.join(output_parent_dirpath, "xml", testcase.name)
-        proto_output_dir_path = os.path.join(output_parent_dirpath, "proto", testcase.name)
+        output_xml_paths: Optional[List[str]] = None
+        output_proto_paths: Optional[List[str]] = None
 
-        os.makedirs(xml_output_dir_path, exist_ok=True)
-        os.makedirs(proto_output_dir_path, exist_ok=True)
+        if testcase.expected_output_xml_path:
```
AsherGlick, line 172: Unless there are multiple expected truthy/falsey values expected here I would rather these be explicit checks for values. Otherwise you will very quickly create undefined behavior when something unexpected evaluates to true or false because it happens to be truthy or falsey when you did not expect it to ever be an option.

xml_converter/integration_tests/run_tests.py
```diff
@@ -166,17 +166,24 @@ def main() -> bool:
...
+            xml_output_dir_path = os.path.join(output_parent_dirpath, "xml", testcase.name)
+            os.makedirs(xml_output_dir_path, exist_ok=True)
+            output_xml_paths = [xml_output_dir_path]
+
+        if testcase.expected_output_proto_path:
```
AsherGlick, line 177: dito the truthy/falsey

AsherGlick (changes requested):
Also make an issue that indicates how to undo the changes you are making to the unit tests that can no longer be tested. Just including a simple git command with the commit hash right before this PR as a way to revert the files is good enough. Someone could of course just search for that hash, or rewrite the tests but might as well have it recorded in an easy to find location.

xml_converter/generators/generate_cpp.py
```diff
@@ -359,7 +362,7 @@ def generate_cpp_variable_data(
                             serialize_proto_side_effects=[],
                             deserialize_proto_function="from_proto_" + component_class_name,
                             deserialize_proto_side_effects=[],
-                        ) if fieldval.protobuf_field is not None and component.protobuf_field is not None else None,
+                        ) if fieldval.protobuf_field is not None and component.protobuf_field is not None and fieldval.exclude_from_protobuf is not True else None,
```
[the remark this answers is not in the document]
klingbolt: You are correct. Since this check prevents the proto_info from being populated, we do not need the 2nd check within the template code.

xml_converter/generators/generate_cpp.py
```diff
@@ -425,7 +428,7 @@ def generate_cpp_variable_data(
                 deserialize_proto_function = fieldval.custom_functions.read_proto
 
             proto_info: Optional[AttributeVariableProtoInfo] = None
-            if fieldval.protobuf_field is not None:
+            if fieldval.protobuf_field is not None and fieldval.exclude_from_protobuf is not True:
```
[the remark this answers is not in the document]
klingbolt: My feeling is it is best to avoid populating the proto_info field if we are skipping that attribute. This will avoid any unforeseen errors in the population functions and prevent any erroneous data from being created.

xml_converter/generators/generate_cpp.py
```diff
@@ -512,15 +516,16 @@ def write_attribute(output_directory: str, data: Dict[str, Document]) -> List[st
         proto_field_type: str = ""
         proto_field_prototype: Optional[str] = None
         for marker_type in attribute_data.applies_to_as_str():
-            new_type = get_proto_field_cpp_type(marker_type, attribute_data.protobuf_field)
-            if proto_field_type != "" and proto_field_type != new_type:
-                print("Proto Field type differes between different marker types for ", attribute_data.protobuf_field)
-            proto_field_type = new_type
+            if attribute_data.exclude_from_protobuf is False:
```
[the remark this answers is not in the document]
klingbolt: agreed, Changed to suggested.

xml_converter/integration_tests/run_tests.py
klingbolt: I think I had an error that required this to be added but after testing it seems that it is no longer needed. I must  have resolved the error elsewhere. The if statements can be removed.

xml_converter/proto/guildpoint.proto
```diff
@@ -21,32 +21,32 @@ message Category {
...
 message Icon {
     uint32 texture_id = 2;
     bytes guid = 3;
     int32 map_id = 4;
-    float distance_fade_end = 5;
-    float distance_fade_start = 6;
-    float height_offset= 7;
+    // float distance_fade_end = 5;
+    // float distance_fade_start = 6;
+    // float height_offset= 7; Double check how this is used
```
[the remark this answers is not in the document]
klingbolt: This was a comment from a discussion we had. I will open an issue that we want to check it and remove the comment here

AsherGlick (approved):

Issue #331, opened 2024-08-18
Standarize normalization of the "type" attribute in xml


AsherGlick:
This should likely be lowercased when we read it. However there may be other normalization we should implement.

PR #367, opened 2024-12-09
Added an option to split the output proto files by top level category

klingbolt:


xml_converter/integration_tests/test_cases/xml_split_by_category/input/pack/xml_file.xml
```diff
@@ -0,0 +1,13 @@
+<OverlayData>
+    <MarkerCategory DisplayName="TopLevel1" Name="toplevel1" XPos="169.81" YPos="210.65" ZPos="215.83" MapID="50" ></MarkerCategory>
+
+    <MarkerCategory DisplayName="TopLevel2" Name="toplevel2" XPos="169.81" YPos="210.65" ZPos="215.83" MapID="50" ></MarkerCategory>
+
+    <MarkerCategory DisplayName="TopLevel3" Name="toplevel3" XPos="169.81" YPos="210.65" ZPos="215.83" MapID="50" ></MarkerCategory>
```
AsherGlick, line 6: Lets differentiate these by more than just `DisplayName` and `Name`. Otherwise this test does a silly job at not catching the issue of "we write the same top level node but with different display/names for each category". The attributes could also be given to the POIs instead of the categories because XPos YPos ZPos dont make a huge amount of sense as category attributes.

Lets also add some subcategories here too so that we can test that "more than just the top level category gets written". Because we do want the entire category tree.

xml_converter/src/packaging_protobin.cpp
```diff
@@ -235,6 +234,8 @@ void write_protobuf_file(
         }
     }
 
+    StringHierarchy category_filter;
+    category_filter.add_path({}, true);
```
AsherGlick, line 238: Why are you hardcoding the global wildcard here and preventing it from being passed in to the function?

xml_converter/src/packaging_protobin.cpp
```diff
@@ -243,10 +244,47 @@ void write_protobuf_file(
...
+        category_filter.add_path({it->first}, true);
+        _write_protobuf_file(
+            join_file_paths(state.marker_pack_root_directory, it->first + ".guildpoint"),
+            category_filter,
+            marker_categories,
+            category_to_pois,
+            &state);
+    }
+}
+
```
AsherGlick, line 284: And then this function duplicates all of the logic of `write_protobuf_file` in order to access the earlier removed ability to pass in a `category_fiter`.

If the counter argument is the building of the `category_to_pois` dict for each file, aka: performance optimizations. I would necessarily have expected to see benchmarks somewhere in this PR

xml_converter/src/packaging_protobin.cpp
```diff
@@ -268,6 +306,9 @@ void write_protobuf_file_per_map_id(
         }
     }
 
+    StringHierarchy category_filter;
+    category_filter.add_path({}, true);
+
```
AsherGlick, line 311: Ditto
> Why are you hardcoding the global wildcard here and preventing it from being passed in to the function?

xml_converter/src/xml_converter.cpp
```diff
@@ -138,16 +138,20 @@ void write_taco_directory(
...
     }
-    StringHierarchy category_filter;
-    category_filter.add_path({}, true);
-    write_protobuf_file(output_path, category_filter, marker_categories, parsed_pois);
+    if (split_proto_by_category) {
+        write_protobuf_file_per_category(output_path, marker_categories, parsed_pois);
+    }
+    else {
+        write_protobuf_file(output_path, marker_categories, parsed_pois);
+    }
```
AsherGlick, line 154: This logic should probably be nearby the logic for `split by map id`. It is convoluted that split by map id is in the parent function but split by category is in this child function.

xml_converter/src/xml_converter.cpp
```diff
@@ -308,6 +314,10 @@ int main(int argc, char* argv[]) {
             allow_duplicates = true;
             arg_target = nullptr;
         }
+        else if (!strcmp(argv[i], "--split-guildpoint-by-category")) {
+            split_proto_by_category = true;
```
AsherGlick, line 318: What happens if `--output-split-guildpoint-path` and `--split-guildpoint-by-category` are both specified. Does the program act in a way that a user could reasonably expect? I think it wont, because it will only split by map id as the current logic for splitting by map id does not interact with the splitting by category logic.

Maybe this is the wrong argument to be exposing to the user? Think through some user stories for this holistically because we can still alter how other command line flags for xml_converter are presented to the user if we think there are better paradigms to use than what we are currently using.

AsherGlick (changes requested):
A good proof of concept but I think some of the minor design decisions chosen here did not take enough into consideration and need to be reevaluated.

klingbolt:
I agree this PR is incomplete. It is also heavily affected by a bug in type that is addressed in #331. In hindsight I should have marked this as a draft. 
I would like your opinion. This is now the third optional flag if you consider, split by mapid, split by category, and allow duplicates. I think it should be possible to want all three of these at the same time. Could we change the arguments to that there is one filepath argument and then all three of these, or any other writing modifiers, as options? Or should it be split so that every optional writing option is its own filepath.

xml_converter/src/packaging_protobin.cpp
klingbolt: I believe that category_filter is not global in the same sense as the other two large inputs. It is created just before this function is called, and always for a single path. In this sense, it is more similar in scope to the `ProtoWriterState`, which is also created in this function.

xml_converter/src/packaging_protobin.cpp
klingbolt: This function actually duplicates `write_protobuf_file_per_map_id`. Since that is now 3 places, perhaps the creation of `category_to_pois` should be its own function. 

My other thought was that I am already iterating over the `marker_categories` map in order to get the name and would need a second iteration if I were to do it for the category filter first before this function is called.

xml_converter/src/xml_converter.cpp
klingbolt: I have actually considered moving map_id into this function but it would require reworking some other variables so thought it outside this PR scope.

AsherGlick:
I have not given the ux much thought yet. But I would think about "is there a situation where we want to output multiple files". I could think that an automated service that is testing these would want to output multiple types at once, maybe for automatic downloads. And we use that within our test harness, but the harness could be changed.

klingbolt:
Closing as its purpose has been replaced by #399

PR #372, opened 2025-01-16
Adding codegen runscript

AsherGlick:
This is a helper script that can be used to generate code even when not in the directory that contains the generator.


klingbolt (approved):
A nice addition!

PR #378, opened 2025-01-19
Command line parser

klingbolt:
Ignore https://github.com/AsherGlick/Burrito/pull/376 as it didn't make the comments on the proto files.
Changed the parser to create a vector of structs that contain the vector of paths but also information about the path's usage and changes to the way the converter uses it. This should facilitate the seamless addition of new arguments.


xml_converter/src/xml_converter.cpp
```diff
@@ -171,7 +192,15 @@ void process_data(
...
         }
     }
-    if (duplicate_categories.size() > 0 && allow_duplicates == false) {
+
+    bool all_allow_duplicates_false = all_of(path_configs.begin(), path_configs.end(), [](const PathConfig p) {
+        if (p.type == "export") {
+            return p.allow_duplicates == false;
+        }
+        return true;
+    });
```
AsherGlick, line 201: This is a very convoluted block of code:
"We will not ignore duplicates if and only if the allow duplicates flag was not present after an export format"

When it probably should be:
"We will ignore duplicates if the allow duplicates flag was passed in"

The root cause is in the parser logic, only supporting flags by assigning them to `PathConfig`s.

The current implementation also will intentionally silently consume and ignore the allow duplicates flag if it was passed in to an input file. I dont think the user interactions were considered here, please break out the different scenarios from a user perspective and what the user would expect.

xml_converter/src/xml_converter.cpp
```diff
@@ -217,80 +249,72 @@ void process_data(
...
 //
 // Example usage
-//   ./xml_converter --input-taco-paths ../packs/marker_pack --output-split-guildpoint-path ../output_packs
-//   ./xml_converter --input-taco-paths ../packs/* --output-split-guildpoint-path ../output_packs
+//   ./xml_converter --input-taco-paths ../packs/marker_pack --output-guildpoint-path ../output_packs
+//   ./xml_converter --input-taco-paths ../packs/* --output-guildpoint-path ../output_packs --split_by_map_id
+//
+// --import_FORMAT_paths <path> --FLAGS [--import_FORMAT_paths <path> --FLAGS]
+// --export_FORMAT_paths <path> --filter "conditions" --FLAGS
+// [--export_FORMAT_paths <path> --filter "conditions" --FLAGS]
```
AsherGlick, line 262: Are these examples?

xml_converter/src/xml_converter.cpp
```diff
@@ -217,80 +249,72 @@ void process_data(
...
-
-    vector<string>* arg_target = nullptr;
+vector<PathConfig> parse_arguments(int argc, char* argv[]) {
+    vector<PathConfig> path_configs;
+    PathConfig* arg_target = nullptr;
 
     for (int i = 1; i < argc; i++) {
         if (!strcmp(argv[i], "--input-taco-path")) {
-            arg_target = &input_taco_paths;
+            path_configs.emplace_back("import", "xml");
```
AsherGlick, line 270: Why did you choose `emplace_back` here?

AsherGlick (changes requested):

xml_converter/src/xml_converter.cpp
klingbolt: What I wanted to happen here in each `if` statement was to create a `PathConfig` object, populate it with the information provided by the argument so it could be sorted later, and use a reference to that object when looking at the next argument to see if it modifies the object further. Since the last part of that needs the correct location of the object, `&path_configs.back()` was used to get a pointer to the location on the vector. So for the other 2 steps, I found that using emplace creates the object directly into the vector as opposed to using push which would create an object and then move it into the vector. Both would achieve the same effect. Emplace then had the pro of being one line of code versus four if using push_back.

xml_converter/src/xml_converter.cpp
klingbolt: Its the template for the command line when we create new arguments. Would it make sense to label them correctly or do they not belong here?

xml_converter/src/xml_converter.cpp
```diff
@@ -91,6 +104,7 @@ void write_taco_directory(
 }
 
 void write_burrito_directory(
+    PathConfig path_config,
     string output_path,
```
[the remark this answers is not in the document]
klingbolt: Technically what it needs is the `path_config.split_by_` information to know which function to call. I wanted to further explore if there would be a difference if we iterated over the path vector within this function instead.

xml_converter/integration_tests/test_cases/xml_split_by_map_id/input/pack/xml_file.xml
```diff
@@ -0,0 +1,12 @@
+<OverlayData>
+    <MarkerCategory DisplayName="My Category" Name="mycategory">
+    </MarkerCategory>
+
+    <POIs>
+        <POI Type="mycategory" XPos="169.81" YPos="210.65" ZPos="215.83" MapID="0" />
+        <POI Type="mycategory" XPos="169.81" YPos="210.65" ZPos="215.83" MapID="1" />
+        <POI Type="mycategory" XPos="169.81" YPos="210.65" ZPos="215.83" MapID="5" />
+        <POI Type="mycategory" XPos="169.81" YPos="210.65" ZPos="215.83" MapID="2147483647" />
+        <POI Type="mycategory" XPos="169.81" YPos="210.65" ZPos="215.83" MapID="50" />
```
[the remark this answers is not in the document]
AsherGlick: Instead of having each field be `XPos="169.81" YPos="210.65" ZPos="215.83"` have them contain distinguishing values. Right now the only distinguishing value is the map id, which is also the value that is being used to split the markers up.

Because if the map split code had a bug where it was not properly grabbing any attribute other than map id then we would never know.
```
for i in range(len(mapid)):
    write_marker(xpos[0], ...., map_id[i])
```
Then this test would "pass" because xpos[0] == xpos[1] == xpos[2] etc.

Adding variation to the xpox ypos zpos values just cements the uniqueness of each POI and further reinforces the idea that we did in fact split the marker correctly.

xml_converter/integration_tests/run_tests.py
```diff
@@ -34,10 +34,10 @@ def run_xml_converter(
         cmd += ["--input-guildpoint-path"] + input_proto
     if output_proto:
         cmd += ["--output-guildpoint-path"] + output_proto
-    if split_output_proto:
-        cmd += ["--output-split-guildpoint-path"] + [split_output_proto]
     if allow_duplicates:
         cmd += ["--allow-duplicates"]
+    if split_by_map_id:
```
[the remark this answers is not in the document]
AsherGlick: That is a good first thought, but assumes that all xml inputs want one argument, and the same argument. Instead just transform the inputs field from a `Dict[str,str]` to a `Dict[str, ConfigData]` where ConfigData is something like `{"type": "xml", "split_by_category": true}`. This will allow the inputs to be specified in an intuitive manner to anyone reading the test config, and will also allow for test inputs to nearly mirror the new cli flags.

xml_converter/src/xml_converter.cpp
```diff
@@ -32,6 +32,19 @@
 
 using namespace std;
 
+struct PathConfig {
+    string type;  // "import" or "export" or "template"
+    string format;  // "xml" or "guildpoint"
+    vector<string> paths;
```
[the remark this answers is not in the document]
AsherGlick: I can see an argument for allowing wildcards expanding to multiple files to be effected by a single flag. If that is the case I would argue the program should not act any differently for either of the two examples, assuming there are two marker packs `one` and `two` in the `./packs` folder.
```bash
xml_converter --input-taco-paths ./packs/* --allow-duplicates
xml_converter --input-taco-paths ./packs/one --allow-duplicates --input-taco-paths ./packs/two --allow-duplicates
```

Because the program should not act any differently in either case, the cli arg parser should return the same values for each case, otherwise the rest of the program needs to have two different codepaths for accomplishing the same goal.

As such I think both examples should have the cli parser returning two structs, each with one path, each with the flags that effect that path set properly.

xml_converter/src/xml_converter.cpp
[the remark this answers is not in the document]
AsherGlick: A desire to further experiment with code options justifies passing duplicate data into the function. That experimentation should have been done well before you attempt to publish the code. Otherwise it is just immediately technical debt.

xml_converter/src/xml_converter.cpp
AsherGlick: Do they belong here? I think they would be better served if they were visible to the user. Perhaps in a `--help` menu.

xml_converter/src/xml_converter.cpp
AsherGlick: That is reasonable.

xml_converter/src/xml_converter.cpp
```diff
@@ -217,80 +249,72 @@ void process_data(
...
-    vector<string> input_guildpoint_paths;
-    vector<string> output_guildpoint_paths;
-    bool allow_duplicates = false;
-
-    // Typically "~/.local/share/godot/app_userdata/Burrito/protobins" for
-    // converting from xml markerpacks to internal protobuf files.
-    vector<string> output_split_guildpoint_paths;
-
-    vector<string>* arg_target = nullptr;
+vector<PathConfig> parse_arguments(int argc, char* argv[]) {
```
AsherGlick, line 264: This function is complex enough that should have unit tests. Look at other examples of the C++ unit tests, like those for the string hierarchy class.

xml_converter/src/xml_converter.cpp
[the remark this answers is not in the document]
klingbolt: Thats a good point. I should then treat this class or struct as like a string+. Each class having one path string with all the options added on. That would also simplify the code by knowing each class is singleton.

xml_converter/integration_tests/test_cases/xml_split_by_map_id/input/pack/xml_file.xml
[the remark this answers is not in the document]
klingbolt: I have changed it so each has a unique value.

xml_converter/src/xml_converter.cpp
```diff
@@ -32,6 +32,19 @@
 
 using namespace std;
 
+struct PathConfig {
+    string type;  // "import" or "export" or "template"
+    string format;  // "xml" or "guildpoint"
+    vector<string> paths;
+    bool allow_duplicates = false;
+    // bool split_by_category = false;
+    bool split_by_map_id = false;
```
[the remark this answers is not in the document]
klingbolt: I have reworked my loop here so that it gives errors if the arguments are given out of order. This should make it so that if the type is import then this  argument is not changed

xml_converter/src/xml_converter.cpp
klingbolt: I changed the design for the output of the argument parser to a class that contains a vector of PathConfigs and the global variables and made `--allow-duplicates` a global. This way it just needs to check for the one value.

xml_converter/src/xml_converter.cpp
klingbolt: I will move this to a help function in a future PR

xml_converter/src/xml_converter.cpp
klingbolt: Added tests

AsherGlick (changes requested):

xml_converter/src/xml_converter.cpp
```diff
@@ -33,16 +34,16 @@
 using namespace std;
 
 map<string, vector<string>> read_taco_directory(
-    string input_path,
+    PathConfig path_config,
```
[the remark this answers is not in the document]
klingbolt: Fair enough. Reverting to just passing in the string.

AsherGlick (approved):

PR #385, opened 2025-02-05
Proto differ checker

klingbolt:
This PR fixes an annoyance where the diff checker for protobuf files was posting a comment for every commit even when an identical comment already existed. 

This checks the most recent review comment on a file and if it matches the current diff, skips making a new comment.

This should reduce the noise in longer PRs.

See https://github.com/klingbolt/Burrito/pull/19 for tests to show that the code works in expected ways.

### DIFF
--- a/.github/workflows/diff_protobin.yml
+++ b/.github/workflows/diff_protobin.yml
@@ -100,7 +100,7 @@ jobs:
             const timestamp_regex = /\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d+ \+\d{4}/g;
 
             // Check if a similar comment already exists
-            if (most_recent_comment && most_recent_comment.replace(timestamp_regex, '') === comment_body.replace(timestamp_regex, '')) {
+            if (most_recent_comment && most_recent_comment.body.replace(timestamp_regex, '') === comment_body.replace(timestamp_regex, '')) {
               console.log(`Skipping file ${file}, identical comment already exists.`);
               continue;
             }
--- a/.github/workflows/diff_protobin.yml
+++ b/.github/workflows/diff_protobin.yml
@@ -62,12 +62,29 @@ jobs:
           const issue_number = context.issue.number;
           const githubToken = process.env.GITHUB_TOKEN;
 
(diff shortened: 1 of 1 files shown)


.github/workflows/diff_protobin.yml
```diff
@@ -62,12 +62,29 @@ jobs:
           const issue_number = context.issue.number;
           const githubToken = process.env.GITHUB_TOKEN;
 
+          // Get the existing comments.
+          const {data: comments} = await github.rest.pulls.listReviewComments({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            pull_number: issue_number,
+          })
```
AsherGlick, line 70: While semicolons are not strictly required, and even though we dont have js linting in this project, and even if we did it probably would not parse the js inside a yaml workflow file, I would like to still add a semicolon here.

.github/workflows/diff_protobin.yml
```diff
@@ -79,6 +96,15 @@ jobs:
               "</details>",
             ].join("\n");
             
+            // Regular expression to remove timestamps from the diff contents
+            const timestamp_regex = /\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d+ \+\d{4}/g;
```
AsherGlick, line 100: This is a good first thought to get rid of the timestamp differences between the two file's diff output. However the way this regex is implemented it will obscure ALL timestamp differences anywhere in the files, even if the timestamp that is changed is within the content of the proto file, which could cause an incorrect diff to be displayed.

I think we dont really need the two header lines' data, the filenames are basically arbitrary filenames for old/new and the timestamps are not incredibly useful either as the files were just created to be diffed. It looks like the `diff` command supports a flag called `--label` that will allow those headers to be overwritten with another custom value.

```diff
-         diff -u $file.textproto._old             $file.textproto._new             > $file._diff || true
+         diff -u $file.textproto._old --label old $file.textproto._new --label new > $file._diff || true
```

You could get more creative with the label names but that can be a different PR.

AsherGlick (changes requested):

PR #386, opened 2025-02-05
Added types to workflow to add trigger when PR is no longer draft

klingbolt:
This change adds checks to the custom diff job to trigger when the PR is opened, the head is updated (new commit), or the status switches from draft to "ready for review". 

Testing to prove this workflow works can be seen here, https://github.com/klingbolt/Burrito/pull/20

### DIFF
--- a/.github/workflows/diff_protobin.yml
+++ b/.github/workflows/diff_protobin.yml
@@ -5,6 +5,7 @@ on:
     paths:
       - '**/*.bin'
       - '**/*.guildpoint'
+    types: [opened, synchronize, ready_for_review]
 jobs:
   custom-diff:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/diff_protobin.yml
```diff
@@ -5,6 +5,7 @@ on:
     paths:
       - '**/*.bin'
       - '**/*.guildpoint'
+    types: [opened, synchronize, ready_for_review]
```
AsherGlick, line 8: Lets add a description of what this is doing, ie how is it changing the default values. Maybe even a link to the docs for it in case the default value changes in the future.

AsherGlick (changes requested):

AsherGlick (approved):

PR #396, opened 2025-02-12
Reading paths after input or output is called. 

klingbolt:
Changed how paths are found in the parser to make sure they can only follow input and output arguments.

The old code assumed any unknown string was a file path.

### DIFF
--- a/xml_converter/tests/test_argument_parser.cpp
+++ b/xml_converter/tests/test_argument_parser.cpp
@@ -151,3 +151,42 @@ TEST_F(ParseArgumentsTest, InvalidNoPathAfterOutput){
     EXPECT_TRUE(parsed_arguments.marker_pack_configs.empty());
     EXPECT_NE(std_err.find("Error: Expected a path to a directory after --output-taco-path"), std::string::npos);
 }
+
+TEST_F(ParseArgumentsTest, InvalidFileAfterWrongArgument){
+    char* argv[] = {
+        (char*)"./xml_converter",
+        (char*)"--input-taco-path",
+        (char*)"input1",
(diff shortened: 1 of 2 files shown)


xml_converter/tests/test_argument_parser.cpp
```diff
@@ -151,3 +151,42 @@ TEST_F(ParseArgumentsTest, InvalidNoPathAfterOutput){
...
+    };
+    int argc = sizeof(argv) / sizeof(char*);
+
+    testing::internal::CaptureStderr();
+    ParsedArguments parsed_arguments = parse_arguments(argc, argv);
+    std::string std_err = testing::internal::GetCapturedStderr();
+
+    EXPECT_TRUE(parsed_arguments.marker_pack_configs.empty());
+    EXPECT_NE(std_err.find("Error: Unknown argument --ERROR"), std::string::npos);
+}
```
AsherGlick, line 192: Can we add one more test that triggers the `Error: Expected a path to a directory after` error?

AsherGlick (changes requested):

xml_converter/tests/test_argument_parser.cpp
klingbolt: I believe those are already covered in the tests, InvalidNoPathAfterInput and InvalidNoPathAfterOutput

AsherGlick (approved):

xml_converter/tests/test_argument_parser.cpp
klingbolt: After testing, I found that the new error is covering every case where the error on old line 61 and 91 were being called.

PR #397, opened 2025-02-13
Added an argument to the parser for splitting by category

klingbolt:
This argument will be used to determine the depth in the category tree that the outputs should be split by with a default value of 0.

### DIFF
--- a/xml_converter/src/argument_parser.cpp
+++ b/xml_converter/src/argument_parser.cpp
@@ -62,7 +62,7 @@ ParsedArguments parse_arguments(int argc, char* argv[]) {
             type = it->second.type;
             format = it->second.format;
             // All flags must be set to default value
-            split_by_category_depth.reset();
+            split_by_category_depth.set_null();
             split_by_map_id = false;
 
             while (i + 1 < argc && string(argv[i + 1]).find("--") != 0) {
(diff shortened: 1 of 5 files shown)


xml_converter/src/int_helper.hpp
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    bool has_value() {
+        return !is_null;
+    }
+
+    void reset() {
+        is_null = true;
+        value = 0;
+    }
```
AsherGlick, line 31: Lets change this to something like `set_none()` instead

xml_converter/src/argument_parser.cpp
```diff
@@ -74,6 +76,22 @@ ParsedArguments parse_arguments(int argc, char* argv[]) {
...
+            split_by_category_depth.set_value(0);
+            while (i + 1 < argc && string(argv[i + 1]).find("--") != 0) {
+                if (is_string_valid_integer(argv[i + 1])) {
+                    split_by_category_depth.set_value(std::stoi(argv[++i]));
+                }
+                else {
+                    cerr << "Error: expected an integer after --split-by-category but received " << argv[i + 1] << endl;
+                    return {};
+                }
+            }
```
AsherGlick, line 93: What happens if we have two arguments after `--split-by-category` instead of just 0 or 1.

Add a new unit test for this as well.

AsherGlick (changes requested):

AsherGlick (approved):

PR #399, opened 2025-02-15
Added function to print outputs in proto by category

klingbolt:


xml_converter/src/packaging_protobin.cpp
```diff
@@ -214,14 +214,8 @@ void _write_protobuf_file(
     outfile.close();
 }
 
-void write_protobuf_file(
-    const string& marker_pack_root_directory,
-    const StringHierarchy& category_filter,
-    const map<string, Category>* marker_categories,
-    const vector<Parseable*>* parsed_pois) {
+std::map<string, std::vector<Parseable*>> _construct_category_to_pois_map(const vector<Parseable*>* parsed_pois) {
```
AsherGlick, line 217: I dont think this function needs an underscore prefix just for being a non-public function.  It would in python but this is C++.

xml_converter/src/packaging_protobin.cpp
```diff
@@ -239,6 +233,19 @@ void write_protobuf_file(
...
+void write_protobuf_file(
+    const string& marker_pack_root_directory,
+    const StringHierarchy& category_filter,
+    const map<string, Category>* marker_categories,
+    const vector<Parseable*>* parsed_pois) {
+    std::map<string, std::vector<Parseable*>> category_to_pois;
+    ProtoWriterState state;
+    state.marker_pack_root_directory = marker_pack_root_directory;
+
+    category_to_pois = _construct_category_to_pois_map(parsed_pois);
```
AsherGlick, line 248: You can create an instantiate `categories_to_pois` in one line, no need to split it up across other declarations.

xml_converter/src/packaging_protobin.cpp
```diff
@@ -284,3 +291,47 @@ void write_protobuf_file_per_map_id(
             &state);
     }
 }
+void _category_filter_by_depth(
```
AsherGlick, line 294: I dont think this function needs an underscore prefix just for being a non-public function.

xml_converter/src/packaging_protobin.cpp
```diff
@@ -284,3 +291,47 @@ void write_protobuf_file_per_map_id(
...
+        }
+
+        _category_filter_by_depth(&(it->second.children), target_depth, category_hierarchy, child_name, current_depth + 1);
+    }
+}
+
+// Write protobuf per category
+void write_protobuf_file_per_category(
+    const string& marker_pack_root_directory,
+    const OptionalInt& split_by_category_depth,
```
AsherGlick, line 316: This should probably just be an int. There is no reason this function should be called with a non-int value, and your code is already assuming that this is an int by calling `.get_value()` without checking if it has a value or not.

xml_converter/src/packaging_protobin.cpp
```diff
@@ -284,3 +291,47 @@ void write_protobuf_file_per_map_id(
...
+    const map<string, Category>* marker_categories,
+    const vector<Parseable*>* parsed_pois) {
+    ProtoWriterState state;
+    state.marker_pack_root_directory = marker_pack_root_directory;
+
+    map<string, StringHierarchy> category_hierarchies;
+    _category_filter_by_depth(marker_categories, split_by_category_depth.get_value(), &category_hierarchies, {});
```
AsherGlick, line 323: I feel like this line should read:
```cpp
map<string, StringHierarchy> category_hierarchies = category_filter_by_depth(marker_categories, split_by_category_depth.get_value());
```

It is generally more readable to return a value than ask for it to be passed in as a pointer, though passing it in as a pointer is done often in c, and unavoidable when writing functions intended to mutate a class outside of the class methods.

In this case I recommend following the pattern setup where `category_filter_by_depth` handles all the external setup and internally calls `_category_filter_by_depth`.

xml_converter/src/packaging_protobin.cpp
```diff
@@ -284,3 +291,47 @@ void write_protobuf_file_per_map_id(
...
+        vector<string> child_name = parent_name;
+        child_name.push_back(it->first);
+        if (target_depth == current_depth) {
+            StringHierarchy category_filter;
+            category_filter.add_path(child_name, true);
+            (*category_hierarchy)[join(child_name, "_")] = category_filter;
+        }
+
+        _category_filter_by_depth(&(it->second.children), target_depth, category_hierarchy, child_name, current_depth + 1);
+    }
```
AsherGlick, line 310: You have no terminal case here to exit the loop when the correct depth is hit, meaning that this will always recurse over the entire node tree regardless of the depth, but only act on the single depth `target_depth`. The `if` should probably contain a `continue` or the recursion should be in an else, or something similar.

xml_converter/src/packaging_protobin.cpp
```diff
@@ -284,3 +291,47 @@ void write_protobuf_file_per_map_id(
...
+    map<string, StringHierarchy>* category_hierarchy,
+    vector<string> parent_name,
+    int current_depth = 0) {
+    for (auto it = marker_categories->begin(); it != marker_categories->end(); it++) {
+        vector<string> child_name = parent_name;
+        child_name.push_back(it->first);
+        if (target_depth == current_depth) {
+            StringHierarchy category_filter;
+            category_filter.add_path(child_name, true);
+            (*category_hierarchy)[join(child_name, "_")] = category_filter;
```
AsherGlick, line 306: If `it->first` can contain an `_` then this could cause a collision in category_hierarchy. If we had a structure like
```
one_two
  three
one
  two_three
```

That would be a valid tree with no duplicates. But both of those would end up producing the key `one_two_three` here and collide.

AsherGlick, line 306: Is this an intentional strategy to overwrite the contents of the map element on a collision?
If not you can do 

```cpp
(*category_hierarchy)[<key>].add_path(child_name, true);
```

because C++ Maps will automatically create the elements if you try to read from that element.

xml_converter/src/xml_converter.cpp
```diff
@@ -104,9 +105,16 @@ void write_burrito_directory(
     }
     StringHierarchy category_filter;
     category_filter.add_path({}, true);
+    if (split_by_map_id && split_by_category_depth.has_value()) {
+        cerr << "Error: --split-by-category and --split-by-map-id can not be called at the same time" << endl;
```
AsherGlick, line 109: Theoretically this might not be a command line arg here. 

```suggestion
        cerr << "Error: 'split by category' and 'split by map id' can't be used at the same time" << endl;
```

AsherGlick (changes requested):

xml_converter/src/packaging_protobin.cpp
klingbolt: I had just defaulted to snake case. I'm going to change this to `.` since i know that should not be in the name field since we are using it as delimiter to determine hierarchy. 

The name field is normalized when it is written proto->xml so perhaps we should look at normalizing the xml->xml as well.

xml_converter/src/packaging_protobin.cpp
klingbolt: I did in another function. I'll make this change.

xml_converter/src/packaging_protobin.cpp
klingbolt: I will add continue. No need to check the children if the depth is reached.

xml_converter/src/packaging_protobin.cpp
klingbolt: Good point. Since I check for a value before this function is called, it should always be an int.

xml_converter/src/packaging_protobin.cpp
klingbolt: I like that pattern. I'll make the change.

xml_converter/src/xml_converter.cpp
klingbolt: Added

xml_converter/src/packaging_protobin.cpp
klingbolt: I had written out the steps so I was sure what was happening. This is a good consolidation.

xml_converter/src/packaging_protobin.cpp
klingbolt: I see. In C++ its more used for helper functions for functions of that name, correct?

xml_converter/integration_tests/test_cases/split_proto_by_category_children/testcase.yaml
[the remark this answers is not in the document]
klingbolt: done

xml_converter/integration_tests/test_cases/split_proto_by_category_children/input/pack/xml_file.xml
```diff
@@ -0,0 +1,62 @@
...
+    <POI Type="parent1.child3.grandchild3" />
+    <POI Type="parent2.child1.grandchild1" />
+    <POI Type="parent2.child1.grandchild2" />
+    <POI Type="parent2.child1.grandchild3" />
+    <POI Type="parent2.child2.grandchild1" />
+    <POI Type="parent2.child2.grandchild2" />
+    <POI Type="parent2.child2.grandchild3" />
+    <POI Type="parent2.child3.grandchild1" />
+    <POI Type="parent2.child3.grandchild2" />
+    <POI Type="parent2.child3.grandchild3" />
```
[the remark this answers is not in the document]
klingbolt: I added a POI to all possible categories for completion. There was no issue. When I call `.add_path(child_name, true)`, `child_name` is a vector so it is adding its parent to the hierarchy as well.

xml_converter/integration_tests/test_cases/split_proto_by_category_children/input/pack/xml_file.xml
```diff
@@ -0,0 +1,62 @@
+<OverlayData>
+    <MarkerCategory DisplayName="My Category" Name="mycategory">
+    </MarkerCategory>
```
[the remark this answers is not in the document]
klingbolt: You are correct. Removing.

xml_converter/integration_tests/test_cases/split_proto_by_category/testcase.yaml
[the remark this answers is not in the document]
klingbolt: I like that change. I was uncertain on the name since what it is testing is how the output looks but I think sticking with the prefix of the input makes sense.

xml_converter/integration_tests/test_cases/split_proto_by_category/input/pack/xml_file.xml
```diff
@@ -0,0 +1,62 @@
...
+    <POI Type="parent1.child3.grandchild3" />
+    <POI Type="parent2.child1.grandchild1" />
+    <POI Type="parent2.child1.grandchild2" />
+    <POI Type="parent2.child1.grandchild3" />
+    <POI Type="parent2.child2.grandchild1" />
+    <POI Type="parent2.child2.grandchild2" />
+    <POI Type="parent2.child2.grandchild3" />
+    <POI Type="parent2.child3.grandchild1" />
+    <POI Type="parent2.child3.grandchild2" />
+    <POI Type="parent2.child3.grandchild3" />
```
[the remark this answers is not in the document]
klingbolt: done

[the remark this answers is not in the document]
klingbolt: done

xml_converter/integration_tests/test_cases/split_proto_by_category/input/pack/xml_file.xml
```diff
@@ -0,0 +1,62 @@
+<OverlayData>
+    <MarkerCategory DisplayName="My Category" Name="mycategory">
+    </MarkerCategory>
```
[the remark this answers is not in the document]
klingbolt: removed

xml_converter/integration_tests/test_cases/split_invlaid/testcase.yaml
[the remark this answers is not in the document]
klingbolt: It should be a unit test. This will require setting up a new file and test so it will be added as a new PR.

xml_converter/src/packaging_protobin.cpp
AsherGlick: Correct, a good example of how I use it is for recursive functions that need to accept more than I want to expose to a user in order to recurse.

xml_converter/src/packaging_protobin.hpp
```diff
@@ -9,6 +9,7 @@
 #include "category_gen.hpp"
 #include "file_helper.hpp"
 #include "guildpoint.pb.h"
+#include "int_helper.hpp"
```
AsherGlick, line 12: Do you need this include anymore?

AsherGlick (changes requested):

AsherGlick (approved):

PR #400, opened 2025-02-16
Namespacing all of the attribute module functions

AsherGlick:
This is a very large change, but a simple one. It adds a namespace to each of the attribute functions of
```
Attribute::<attributetype>>::
```
Using the attribute as part of the function's name it also renames all the conversion functions to follow the pattern:
```
from_xml_attribute()
to_xml_attribute()
from_proto_field()
to_proto_field()
```
which both simplifies and unifies all of the function names. Furthermore it makes it clear that any new functions to convert an attribute to/from an external type should follow that format. Eg: `to_json_field` and `from_json_field` are obvious deductions (hopefully).

This seems generally like a good idea for organization and readability but I am open to feedback.


klingbolt:
I like this change. Should make it very easy to read how the functions work.

klingbolt (approved):

PR #427, opened 2025-03-15
Duplicates tracking ID instead of name

klingbolt:


xml_converter/src/xml_converter.cpp
```diff
@@ -33,46 +34,58 @@
...
     string directory_name = filesystem::path(input_path).filename();
     vector<MarkerPackFile> xml_files = get_files_by_suffix(input_path, ".xml");
     for (const MarkerPackFile& path : xml_files) {
-        set<string> top_level_category_names = parse_xml_file(path, marker_categories, parsed_pois);
+        map<Attribute::UniqueId::UniqueId, Category*> top_level_categories = parse_xml_file(path, marker_categories, parsed_pois);
         string file_path = join_file_paths(input_path, path.relative_filepath);
-        for (set<string>::iterator it = top_level_category_names.begin(); it != top_level_category_names.end(); it++) {
-            top_level_category_file_locations[*it].push_back(file_path);
+        for (map<Attribute::UniqueId::UniqueId, Category*>::iterator it = top_level_categories.begin(); it != top_level_categories.end(); it++) {
+            top_level_category_file_locations[it->first].category = it->second;
```
AsherGlick, line 63: The existing an new pointers should be the same, should we be sanity checking that here?
It will require a little bit of extra logic because the first iteration will have garbage in `.category` so we can't compare on every iteration.
I worry mostly about a possible edge case where the pointer is not the same for some reason and it gets overwritten causing unknown behavior.

xml_converter/src/xml_converter.cpp
```diff
@@ -33,46 +34,58 @@
...
     string directory_name = filesystem::path(input_path).filename();
     vector<MarkerPackFile> burrito_files = get_files_by_suffix(input_path, ".guildpoint");
     for (const MarkerPackFile& path : burrito_files) {
-        set<string> top_level_category_names = read_protobuf_file(path, marker_categories, parsed_pois);
+        map<Attribute::UniqueId::UniqueId, Category*> top_level_categories = read_protobuf_file(path, marker_categories, parsed_pois);
         string file_path = join_file_paths(input_path, path.relative_filepath);
-        for (set<string>::iterator it = top_level_category_names.begin(); it != top_level_category_names.end(); it++) {
-            top_level_category_file_locations[*it].push_back(file_path);
+        for (map<Attribute::UniqueId::UniqueId, Category*>::iterator it = top_level_categories.begin(); it != top_level_categories.end(); it++) {
+            top_level_category_file_locations[it->first].category = it->second;
```
AsherGlick, line 87: ditto sanity check

xml_converter/src/xml_converter.cpp
```diff
@@ -146,13 +159,14 @@ void process_data(ParsedArguments parsed_arguments) {
...
-        map<string, vector<string>> top_level_category_file_locations = read_taco_directory(
+        map<Attribute::UniqueId::UniqueId, CategoryWithinSinglePack> top_level_category_file_locations = read_taco_directory(
             marker_pack_config[i].path,
             &marker_categories,
             &parsed_pois
         );
-        for (map<string, vector<string>>::iterator it = top_level_category_file_locations.begin(); it != top_level_category_file_locations.end(); it++) {
-            top_level_category_file_locations_by_pack[it->first].push_back(it->second);
+        for (map<Attribute::UniqueId::UniqueId, CategoryWithinSinglePack>::iterator it = top_level_category_file_locations.begin(); it != top_level_category_file_locations.end(); it++) {
+            top_level_category_file_locations_by_pack[it->first].category = it->second.category;
```
AsherGlick, line 168: ditto sanity check

xml_converter/src/xml_converter.cpp
```diff
@@ -168,25 +182,31 @@ void process_data(ParsedArguments parsed_arguments) {
...
-        map<string, vector<string>> top_level_category_file_locations = read_burrito_directory(
+        map<Attribute::UniqueId::UniqueId, CategoryWithinSinglePack> top_level_category_file_locations = read_burrito_directory(
             marker_pack_config[i].path,
             &marker_categories,
             &parsed_pois
         );
-        for (map<string, vector<string>>::iterator it = top_level_category_file_locations.begin(); it != top_level_category_file_locations.end(); it++) {
-            top_level_category_file_locations_by_pack[it->first].push_back(it->second);
+        for (map<Attribute::UniqueId::UniqueId, CategoryWithinSinglePack>::iterator it = top_level_category_file_locations.begin(); it != top_level_category_file_locations.end(); it++) {
+            top_level_category_file_locations_by_pack[it->first].category = it->second.category;
```
AsherGlick, line 191: ditto sanity check

xml_converter/src/xml_converter.cpp
```diff
@@ -32,47 +33,83 @@
...
     for (const MarkerPackFile& path : xml_files) {
-        set<string> top_level_category_names = parse_xml_file(path, marker_categories, parsed_pois);
+        map<UniqueId, Category*> top_level_categories = parse_xml_file(path, marker_categories, parsed_pois);
         string file_path = join_file_paths(input_path, path.relative_filepath);
-        for (set<string>::iterator it = top_level_category_names.begin(); it != top_level_category_names.end(); it++) {
-            top_level_category_file_locations[*it].push_back(file_path);
+        for (map<UniqueId, Category*>::iterator it = top_level_categories.begin(); it != top_level_categories.end(); it++) {
+            if (top_level_category_file_locations.find(it->first) != top_level_category_file_locations.end()) {
+                if (top_level_category_file_locations[it->first].category != it->second) {
+                    cerr << "Error: Different MarkerCategory objects were found with same ID" << endl;
```
AsherGlick, line 65: Can we add a testcase that triggers this error?

xml_converter/src/xml_converter.cpp
```diff
@@ -32,47 +33,83 @@
...
     for (const MarkerPackFile& path : burrito_files) {
-        set<string> top_level_category_names = read_protobuf_file(path, marker_categories, parsed_pois);
+        map<UniqueId, Category*> top_level_categories = read_protobuf_file(path, marker_categories, parsed_pois);
         string file_path = join_file_paths(input_path, path.relative_filepath);
-        for (set<string>::iterator it = top_level_category_names.begin(); it != top_level_category_names.end(); it++) {
-            top_level_category_file_locations[*it].push_back(file_path);
+        for (map<UniqueId, Category*>::iterator it = top_level_categories.begin(); it != top_level_categories.end(); it++) {
+            if (top_level_category_file_locations.find(it->first) != top_level_category_file_locations.end()) {
+                if (top_level_category_file_locations[it->first].category != it->second) {
+                    cerr << "Error: Different MarkerCategory objects were found with same ID" << endl;
```
AsherGlick, line 101: Can we add a testcase that triggers this error?

xml_converter/src/xml_converter.cpp
```diff
@@ -168,27 +205,23 @@ void process_data(ParsedArguments parsed_arguments) {
...
 
-    for (map<string, vector<vector<string>>>::iterator it = top_level_category_file_locations_by_pack.begin(); it != top_level_category_file_locations_by_pack.end(); it++) {
-        if (it->second.size() != 1) {
-            for (size_t i = 0; i < it->second.size(); i++) {
-                for (size_t j = 0; j < it->second[i].size(); j++) {
-                    duplicate_categories[it->first].insert(it->second[i][j]);
-                }
-            }
+    for (map<UniqueId, CategoryWithinMultiplePacks>::iterator it = top_level_category_file_locations_by_pack.begin(); it != top_level_category_file_locations_by_pack.end(); it++) {
+        if (it->second.categories.size() != 1) {
```
AsherGlick, line 223: This should be `> 1` not `!= 1`. While not currently possible in the code, this `if` would trigger if `size()` were somehow `0` (maybe some sort of postprocessing filtering is done after the fact in the future). A `size_t` is unsigned and cant be negative so no need to worry about negative numbers.

If you are concerned about `0` it would be wise to add a sanity check `else if` to make sure that the `size()` is not 0, and have a unique error in that situation.

AsherGlick (changes requested):

xml_converter/integration_tests/test_cases/proto_and_xml_input_ID_conflicts/input/pack/xml_file.xml
AsherGlick, line 1: Lets split up this test into two tests. This is actually testing if an ID in a single marker pack category has two category nodes assigned to it with different names.

The letters ID should be lowercase if they are used in the new tests.

xml_converter/src/xml_converter.cpp
```diff
@@ -32,47 +33,121 @@
...
+                if (top_level_category_file_locations[it->first].category != it->second) {
+                    id_conflicts[it->first][top_level_category_file_locations[it->first].category->name] = top_level_category_file_locations[it->first].file_paths;
+                    id_conflicts[it->first][it->second->name].push_back(file_path);
+                    continue;
+                }
+            }
+            else {
+                top_level_category_file_locations[it->first].category = it->second;
+            }
+            top_level_category_file_locations[it->first].file_paths.push_back(file_path);
```
AsherGlick, line 73: I think this logic is a little convoluted and needs to be simplified as we discussed offline.

xml_converter/src/xml_converter.cpp
```diff
@@ -32,47 +33,121 @@
...
     }
 
+    if (id_conflicts.size() > 0) {
+        for (map<UniqueId, map<string, vector<string>>>::iterator id_conflicts_it = id_conflicts.begin(); id_conflicts_it != id_conflicts.end(); id_conflicts_it++) {
+            cerr << "Error: Different MarkerCategory nodes were found with same ID (" << id_conflicts_it->first.get_value() << ")" << endl;
+            map<string, vector<string>> category_name_to_file_names;
+            for (map<string, vector<string>>::iterator category_it = id_conflicts_it->second.begin(); category_it != id_conflicts_it->second.end(); category_it++) {
+                for (size_t i = 0; i < category_it->second.size(); i++) {
+                    if (category_it->first == "") {
+                        category_name_to_file_names["not found"].push_back(category_it->second[i]);
```
AsherGlick, line 84: Is this a magic string that we are using here of "not found". Is it possible for a real value to be "not found"?

xml_converter/src/xml_converter.cpp
```diff
@@ -32,47 +33,121 @@
...
+                    }
+                    else {
+                        category_name_to_file_names["\"" + category_it->first + "\""].push_back(category_it->second[i]);
+                    }
+                }
```
AsherGlick, line 143: Ditto on the previous comments in the similar block of code.

xml_converter/src/xml_converter.cpp
```diff
@@ -199,11 +270,25 @@ void process_data(ParsedArguments parsed_arguments) {
...
-            for (string str : it->second) {
-                cout << "        " << str << endl;
+        cout << "The following top level categories had a conflict in IDs. For XML, these IDs may be generated from the 'name' attribute." << endl;
+        for (map<UniqueId, CategoryWithinMultiplePacks>::iterator it = duplicate_categories.begin(); it != duplicate_categories.end(); it++) {
+            cout << "    Categories were found that share the ID \"" << it->first.get_value() << "\"" << endl;
+            map<string, vector<string>> category_name_to_file_names;
+            for (CategoryWithinSinglePack category_within_single_pack : it->second.categories) {
+                for (size_t i = 0; i < category_within_single_pack.file_paths.size(); i++) {
+                    if (category_within_single_pack.category->name == "") {
+                        category_name_to_file_names["not found"].push_back(category_within_single_pack.file_paths[i]);
```
AsherGlick, line 280: Is this a magic word.

AsherGlick (changes requested):

AsherGlick (changes requested):

identify_duplicate_categories.gd
```diff
@@ -21,37 +21,46 @@
...
 				# Remove the first 12 spaces
-				var file_name = line.substr(8)
-				collisions[category].append(file_name)
+				var file_name = line.substr(12)
+				collisions[id][category].append(file_name)
 			elif line.begins_with("        In these files the Category 'name' is "):
-				# Remove the first 38 characters and the open quote
-				var line_without_prefix = line.substr(39)
+				# Remove the first 46 characters and the open quote
+				var line_without_prefix = line.substr(47)
```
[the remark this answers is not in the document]
klingbolt: I'm adding a check that if there is no end quote, the category name is `""`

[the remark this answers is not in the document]
AsherGlick: This seems like quite a hacky solution. Lets come up with something a little less fragile to change here. An aside: I feel all of these corner cases stem from the use of the magic string values.

AsherGlick (changes requested):

AsherGlick (approved):

PR #432, opened 2025-03-26
Change to default checked state

klingbolt:
Changed how the default for checked worked to make it easier to view full marker packs

### DIFF
--- a/Spatial.gd
+++ b/Spatial.gd
@@ -543,10 +543,10 @@ func init_category_tree():
 
 func guildpoint_categories_to_godot_nodes():
 	for guildpoint_category in self.guildpoint_data.get_category():
-		_guildpoint_categories_to_godot_nodes(null, guildpoint_category, self.markers_3d, self.markers_2d, false)
+		_guildpoint_categories_to_godot_nodes(null, guildpoint_category, self.markers_3d, self.markers_2d, 0)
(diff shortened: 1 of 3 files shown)


Spatial.gd
```diff
@@ -565,9 +565,15 @@ func _guildpoint_categories_to_godot_nodes(item: TreeItem, guildpoint_category:
...
-	category_item.set_checked(1, Settings.local_category_data.get(Marshalls.raw_to_base64(category_data.guildpoint_category.get_id()), {}).get("checked", false))
+	var checked = true
+	var collapsed = true
+	if depth in [0, 1]:
+		checked = false
+	category_item.set_checked(1, Settings.local_category_data.get(Marshalls.raw_to_base64(category_data.guildpoint_category.get_id()), {}).get("checked", checked))
 	category_item.set_tooltip(1, "Show/Hide")
 	category_item.set_editable(1, true)
+	if depth == 0:
+		collapsed = false
```
AsherGlick, line 576: Lets move this closer to the instantiation of collapsed, or the collapsed instantiation closer to this. There is a weird nebulous zone between 569 and 575 where collapsed will be `true` even if `depth == 0`

AsherGlick (changes requested):

PR #442, opened 2025-04-16
Added filters to the file selection window

klingbolt:
Adding these filters makes it so files that should not be selected are not shown. Should make the user experience smoother. 

Without filter 
image
With filter 
image


AsherGlick:
Rebase this on top of your other PR to add apt get update to the CI?

AsherGlick:
Can you attach a screenshot of what the modified file selectors UI looks like before and after this change?

PR #444, opened 2025-04-17
Add setting to override Burrito icon position.

aukuste:
This adds a toggle button and two text input fields that allows you to exactly specify the position of the Burrito icon.

I needed this because I also use Raidcore Nexus, which puts the icon in the exact same place as Burrito does, causing me to be unable to click on the Raidcore Nexus icon. This pull request aims to fix this issue by introducing a new setting that allows you to specify the position of the Burrito icon.

Now, regarding the changes to the .import files, I honestly have no clue why this happened and this wasn't a conscious change on my part. Is that something that I should revert?

The removal of "res://addons/protobuf/plugin.cfg" from the editor_plugins part in the project.godot file seems to happen every time I start Godot because the file doesn't exist in the repository, so I left that change in there.

Here's what the settings menu looks like now:
Screenshot_20250417_144408

If there's anything that I should change with this pull request then please tell me and I'll change whatever is necessary. I'm unsure if this was the best way to implement this, but I'll try to explain my reasoning for some of the changes:

The introduction of GlobalMenuButton.gd:
This allows for realtime feedback when you change the position of the icon, instead of kind of randomly whenever the icon position is updated whenever Burrito Link (I think it's this anyway) sends an update or whatever.

The part where the tooltip text for the toggle button is set in Spatial.tscn:
Honestly not sure where I should've put this, but I feel like this is important so that you can perfectly pixel align the positioning by being able to see the default position data. Should I perhaps include what UI size the user is on in this tooltip? I don't feel like it's necessary tbh.


Spatial.gd
```diff
@@ -364,8 +373,12 @@ func decode_context_packet(spb: StreamPeerBuffer):
...
 		self.ui_size = 1
 
-	$Control/GlobalMenuButton.margin_left = self.button_margin[self.ui_size]["left"]
-	$Control/GlobalMenuButton.margin_right = self.button_margin[self.ui_size]["right"]
+	if Settings.override_burrito_icon_position_enabled == true:
+		$Control/GlobalMenuButton.set_position(Vector2(Settings.override_burrito_icon_horizontal_position, Settings.override_burrito_icon_vertical_position))
+	else:
+		$Control/GlobalMenuButton.margin_left = self.button_margin[self.ui_size]["left"]
+		$Control/GlobalMenuButton.margin_right = self.button_margin[self.ui_size]["right"]
+
```
AsherGlick, line 381: I think this `else` block should be moved to be within `GlobalMenuButton.gd` and this line should just call `GlobalMenuButton.set_position(...)`. Having one place that controls the position of the UI node makes it easier to understand. Also, with this code here if the user disabled the position override, will the icon return to its default position?

Spatial.gd
```diff
@@ -99,6 +99,15 @@ func _ready():
...
 		OS.window_size = Vector2(Settings.minimum_width, Settings.minimum_height)
+
+	$Control/Dialogs/SettingsDialog/ScrollContainer/GridContainer/OverrideBurritoIconPosition.hint_tooltip = "Default horizontal positions:\n" \
++ "Small: {small}\n".format({"small": button_margin[0]["left"]}) \
++ "Normal: {normal}\n".format({"normal": button_margin[1]["left"]}) \
++ "Large: {large}\n".format({"large": button_margin[2]["left"]}) \
++ "Larger: {larger}\n".format({"larger": button_margin[3]["left"]}) \
++ "\n" \
++ "Width of icon: {icon_width}".format({"icon_width": $Control/GlobalMenuButton/TextureRect.get_texture().get_size().x})
+
```
AsherGlick, line 110: This feels like it should be inside of SettingsDialog.gd. I am guessing it is here because the constant `button_margin` is here and importing that elsewhere is difficult.

Lets move `button_margin` and `minimap_scale` to settings.gd. Then when we need to access that data in this file or in the settings dialog file we can access it via the settings singleton. This way all of the modifications and configurations for the settings dialog scene is contained within the settings dialog file and easy to locate.

Gizmo/gizmo_base.glb.import
```diff
@@ -22,6 +22,8 @@ materials/storage=1
 materials/keep_on_reimport=true
 meshes/compress=true
 meshes/ensure_tangents=true
+meshes/octahedral_compression=true
+meshes/vertex_cache_optimization=true
```
AsherGlick, line 26: I am not sure why godot wanted to update all these .import files. Usually it does not do that when we are all using the same version of godot (which is what the version restrict plugin tries to force). But I dont think any of these changes are relevant to the PR, so lets revert all of the .import changes.

project.godot
```diff
@@ -49,7 +49,7 @@ window/per_pixel_transparency/enabled=true
 
 [editor_plugins]
 
-enabled=PoolStringArray( "res://addons/protobuf/plugin.cfg", "res://addons/version_restrict/plugin.cfg" )
```
AsherGlick, line 52: Lets revert this as well.

@klingbolt can you look into this and see what we can do about either including the plugin (if the licenses are compatible) or removing it from this plugin list.

Spatial.tscn
```diff
@@ -683,123 +685,162 @@ __meta__ = {
...
+margin_top = 200.0
+margin_right = 372.0
+margin_bottom = 224.0
+size_flags_horizontal = 3
+
+[node name="OverrideBurritoIconVerticalPositionLabel" type="Label" parent="Control/Dialogs/SettingsDialog/ScrollContainer/GridContainer"]
+margin_top = 233.0
+margin_right = 192.0
+margin_bottom = 247.0
+text = "Horizontal Icon Position"
```
AsherGlick, line 780:
```suggestion
text = "Vertical Icon Position"
```

Spatial.tscn
```diff
@@ -683,123 +685,162 @@ __meta__ = {
 
 [node name="MinimumWidthLabel" type="Label" parent="Control/Dialogs/SettingsDialog/ScrollContainer/GridContainer"]
 margin_top = 5.0
-margin_right = 112.0
+margin_right = 192.0
```
AsherGlick, line 688: Can you add a before/after image to the pr? I want to see what is causing all the `margin_left` `margin_right` deltas. I am guessing it is just one of the new labels you added being wider than and of the previous labels, which is fine.

AsherGlick (changes requested):
Thanks for the PR! I like the overall concept. I have a few notes on the implementation but otherwise it looks good.

Spatial.gd
aukuste: > I think this else block should be moved to be within GlobalMenuButton.gd and this line should just call GlobalMenuButton.set_position(...). Having one place that controls the position of the UI node makes it easier to understand.

Agreed. This would however require me to move ui_size from Spatial.gd to Settings.gd (because of it being a singleton) however, because ui_size needs to be available globally since it's required for setting the horizontal position of the icon. Would it be fine if I moved ui_size to Settings.gd? Or do you have a better solution in mind?

> Also, with this code here if the user disabled the position override, will the icon return to its default position?

Yes. It wouldn't happen instantly, but it'd happen within about 1-2 seconds after turning off the setting. The decode_context_packet function runs now and then so the position update would happen in a reasonable amount of time at least (I'm not familiar enough with the code base yet, but it's what I've noticed during testing at least).

I've made some changes so that the position of the icon is updated via the _update_global_menu_button_position function in GlobalMenuButton.gd in that part of the code in Spatial.gd, which also results in the position change being instant when you turn the position override setting on and off.

Spatial.gd
aukuste: > This feels like it should be inside of SettingsDialog.gd. I am guessing it is here because the constant `button_margin` is here and importing that elsewhere is difficult.

Yep, that's exactly why.

> Lets move `button_margin` and `minimap_scale` to settings.gd. Then when we need to access that data in this file or in the settings dialog file we can access it via the settings singleton. This way all of the modifications and configurations for the settings dialog scene is contained within the settings dialog file and easy to locate.

I was thinking of doing this, but I wanted to make changes that were as non-disruptive as possible in terms of code changes. But with confirmation from you that it's fine, I've gone ahead and made that change locally. Will commit and push my changes either later today or tomorrow.

Gizmo/gizmo_base.glb.import
aukuste: It may have been because I tried to open the project with Godot 4.4.1 at first, which failed, and informed me to use Godot 3. So I tried loading the project with Godot 3.6, but that didn't work either, but it may have been the cause for these .import changes?

Some info in the README about how to build the project, and what version of Godot to use would probably be a good idea. Wasn't too hard to figure it out eventually, but I was a bit puzzled when the project wouldn't build in Godot :)

aukuste: I'll revert the changes of course, just to make that clear.

Spatial.gd
AsherGlick: > Agreed. This would however require me to move ui_size from Spatial.gd to Settings.gd (because of it being a singleton) however, because ui_size needs to be available globally since it's required for setting the horizontal position of the icon. Would it be fine if I moved ui_size to Settings.gd? Or do you have a better solution in mind?

Lets move ui_size to settings.gd. In most of my projects that have settings files I see "state" objects eventually get moved into the settings module, then if there are too many pieces of state they are broken out. Lets not have ui_size as a saveable setting though, just as a state that happens to live in settings for now.

aukuste:
Settings menu before my changes (based on commit c66019b1e65c5e0bbf08f4bc44bcb3ec8917b986):
Screenshot_20250418_203314

After my changes (as of commit c766600028fa6cc41ad41680ec5ab62bb19d2f66):
Screenshot_20250418_203421

I'm slightly bothered about the scroll bar showing up there, so I may have a look at that. Hopefully that's not too hard to fix.

Spatial.tscn
aukuste: See my comment here: https://github.com/AsherGlick/Burrito/pull/444#issuecomment-2815998414

aukuste:
Are there any other changes that I should make, or does it look good as it is right now?

AsherGlick (approved):

AsherGlick:
Looks good, I am not too concerned about the scroll bar, it was going to happen eventually.

PR #462, opened 2025-04-30
Rename xml_converter to burrito_converter

AsherGlick:
This is a big PR but there are not very many changes. All of this was done via a script that can be found at `scripts/rename_burrito_converter.sh`. This may be useful in understanding what changed at a glance.

As a part of this PR please test the burrito overlay to be sure that it works with importing various marker packs on your machine.


klingbolt (approved):
Script tested and identical output. Build works in game.

PR #470, opened 2025-05-08
Changed how converter path is defined

klingbolt:
This makes the relative path to the converter flexible in case burrito is launched from another directory.

### DIFF
--- a/FileHandler.gd
+++ b/FileHandler.gd
@@ -1,13 +1,13 @@
 const identify_duplicate_categories = preload("res://identify_duplicate_categories.gd")
 
-const converter_executable_path = "./burrito_converter"
-
 static func call_burrito_converter(args: PoolStringArray) -> Dictionary:
 	var output: Array = []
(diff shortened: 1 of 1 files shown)


FileHandler.gd
```diff
@@ -1,13 +1,13 @@
...
-const converter_executable_path = "./burrito_converter"
-
 static func call_burrito_converter(args: PoolStringArray) -> Dictionary:
 	var output: Array = []
-	var path_to_executable: String = converter_executable_path
+	var path_to_executable: String
 	print(args)
 	if OS.has_feature("editor"):
-		path_to_executable += "/build/burrito_converter"
+		path_to_executable = "./burrito_converter/build/burrito_converter"
```
AsherGlick, line 8: Does get executable dir not work for `editor` builds? Is it even possible to trigger an editor build from outside of the projects working directory?

AsherGlick (approved):

PR #501, opened 2025-10-12
Splitting out arguments into individual structs for better control

AsherGlick:
Using structs gives us some better control over which flags can be assigned to which inputs and outputs, allowing us to do some more complex things with arguments in the future.

### DIFF
--- a/burrito_converter/src/argument_parser.cpp
+++ b/burrito_converter/src/argument_parser.cpp
@@ -9,19 +9,25 @@
 
 using namespace std;
 
-InputTacoCommand::InputTacoCommand() : path("") {
+InputTacoCommand::InputTacoCommand()
+    : path("") {
 }
-InputTacoCommand::InputTacoCommand(string path) : path(path) {
-}
-
-InputGuildpointCommand::InputGuildpointCommand() : path("") {
-}
-InputGuildpointCommand::InputGuildpointCommand(string path) : path(path) {
+InputTacoCommand::InputTacoCommand(string path)
+    : path(path) {
 }
 
-OutputTacoCommand::OutputTacoCommand() : path("") {
+InputGuildpointCommand::InputGuildpointCommand()
+    : path("") {
 }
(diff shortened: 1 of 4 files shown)


burrito_converter/src/burrito_converter.cpp
```diff
@@ -316,10 +322,17 @@ void process_data(ParsedArguments parsed_arguments) {
     // Write all of the protobin guildpoint paths
     begin = chrono::high_resolution_clock::now();
     for (size_t i = 0; i < marker_pack_config.size(); i++) {
-        if (marker_pack_config[i].type != BehaviorType::EXPORT || marker_pack_config[i].format != MarkerFormat::GUILDPOINT) {
+        auto* command = dynamic_cast<OutputGuildpointCommand*>(marker_pack_config[i]);
```
klingbolt, line 329: This line should be formatted similar to line 329

burrito_converter/src/argument_parser.cpp
```diff
@@ -3,19 +3,39 @@
...
+OutputGuildpointCommand::OutputGuildpointCommand(
+    std::string path,
+    OptionalInt split_by_category_depth,
+    bool split_by_map_id
+) : path(path),
+    split_by_category_depth(split_by_category_depth),
+    split_by_map_id(split_by_map_id) {
 }
 
 class ArgumentConfig {
```
klingbolt: This class should be removed if it is no longer used.

klingbolt (changes requested):

burrito_converter/src/burrito_converter.cpp
AsherGlick: Was this supposed to be on line 315?
329 is a `write_burrito_directory` call

klingbolt: It was. Not sure how it ended up here.

klingbolt (approved):

PR #509, opened 2025-12-03
Update input region based on visible dialogs

DziubanMaciej:
This commit implements input region setting in burrito-fg instead of relying on Godot's API, which only allows a single rect. It enhances Burrito usability - it's not needed to close all dialogs to interact with the game.

- add `set_optimal_mouse_block`, which iterates over all dialogs and applies their bounding rects as input rects. Only supported for x11 for now. For other backend it will return false.
- add `update_mouse_block` as a frontend function for  `set_minimal_mouse_block`, `set_maximal_mouse_block` and the newly added `set_optimal_mouse_block`. It will first try to use optimal mouse block and if it fails, is not supported or is disabled with a setting, it will fallback to min/max mouse block.
- Connect the `item_rect_changed` and `visibility_changed` of all visible controls to `update_mouse_block` (although not directly, but via a timer). There's no reason to call this function in code.
- `InputRegionUpdateTimer` is added to limit the number of calls to `update_mouse_block`. When any dialog changes visibility or bounding rect, it restarts this timer. When the timer elapses, it calls `update_mouse_block`.

### DIFF
--- a/Settings.gd
+++ b/Settings.gd
@@ -45,6 +45,7 @@ var burrito_link_wine_path: String = ""
 var burrito_link_env_args: String = ""
 
 var enable_player_cutout: bool = false
+var allow_optimal_mouse_block: bool = false
 
 # We save the marker data in this directory when the files are have been split
 # by Map ID. All changes made by the editor are automatically saved in these
@@ -102,6 +103,8 @@ func _ready():
 		self.start_with_open_menu = self._config_data["start_with_open_menu"]
 	if "enable_player_cutout" in self._config_data:
 		self.enable_player_cutout = self._config_data["enable_player_cutout"]
+	if "allow_optimal_mouse_block" in self._config_data:
+		self.allow_optimal_mouse_block = self._config_data["allow_optimal_mouse_block"]
 
 
 func get_saved_markers_dir() -> String:
@@ -127,7 +130,8 @@ func save():
 		"burrito_link_env_args": burrito_link_env_args,
 		"local_category_data": local_category_data,
 		"start_with_open_menu": start_with_open_menu,
-		"enable_player_cutout": enable_player_cutout
+		"enable_player_cutout": enable_player_cutout,
+		"allow_optimal_mouse_block": allow_optimal_mouse_block
 	}
 
 	var file = File.new()
--- a/SettingsDialog.gd
+++ b/SettingsDialog.gd
@@ -29,6 +29,9 @@ func load_settings():
 	var enable_player_cutout: CheckButton = $ScrollContainer/GridContainer/EnablePlayerCutout
 	enable_player_cutout.pressed = Settings.enable_player_cutout
 
+	var allow_optimal_mouse_block: CheckButton = $ScrollContainer/GridContainer/AllowOptimalMouseBlock
+	allow_optimal_mouse_block.pressed = Settings.allow_optimal_mouse_block
+
 
 func save_settings(new_value=null):
 	var minimum_width: LineEdit = $ScrollContainer/GridContainer/MinimumWidth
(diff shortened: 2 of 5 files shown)


AsherGlick:
I have only skimmed the code so far, but I had a question before I begin going through this in depth. Why do we require any of this code to be in burrito-fg's rust code instead of in the core gdscript? gdscript is going to be much more flexible and cross platform. Examples of why we might need to write the logic in rust:
1) Need to call c-api code, which would require an extension no matter what and burrito-fg is already in place
2) If the performance of gdscript is measurably too slow, in which case we should include some sort of profiling data in the pr

DziubanMaciej:
We cannot do this in gdscript, because Godot does not expose a way to set multiple input regions at the same time. `OS.set_window_mouse_passthrough` takes only one polygon. Internally it calls the same APIs as I did here (`XPolygonRegion`, `XShapeCombineRegion`), but each subsequent call fully overwrites what the previous one did.

AsherGlick:
That seems like it would be a solvable problem in code though, or is there something more I am missing? This solution means that we have to be the maintainers of all mouse passthrough implementations for each platform we want to target. If godot is already happy to maintain those implementations using a stable api I dont see a good reason why this project should take on that burden as well.

Even discounting mac and windows, this looks like an x11-only solution with no fallback. Which means at the very least it will need to be rewritten for wayland at some point in the near future or break completely. Not working at all is a worse outcome than the current implementation so I am not inclined to accept this design just because this api allows us to specify multiple input regions. 

I would rather maintain a single implementation that was able to combine multiple bounding rectangles into one overall bounding line, which is not likely to change over time, than maintain multiple mouse passthrough implementations.

DziubanMaciej:
> That seems like it would be a solvable problem in code though, or is there something more I am missing?

How? There's literally no Godot API that would allow this. 

> If godot is already happy to maintain those implementations using a stable api I dont see a good reason why this project should take on that burden as well.

But the API is not sufficient and you had to work around it by capturing entire screen, which is not very pleasant to use. The reason to do this would be usability.

> Even discounting mac and windows, this looks like an x11-only solution with no fallback. Which means at the very least it will need to be rewritten for wayland at some point in the near future or break completely.

I think I've seen multiple people complaining burrito is unusable on Wayland and even you telling people they should use X11. So it could be implemented for Wayland when you decide you want to officially support it. As for windows... I guess everyone just uses blish :D

> I would rather maintain a single implementation that was able to combine multiple bounding rectangles into one overall bounding line, which is not likely to change over time, than maintain multiple mouse passthrough implementations.

But this one overall bounding line will only work on X11 (and maybe windows) as well. Wayland's API only allow rectangles, which is not implemented in Godot. Godot would have to add a new API. There's been a PR for 2 years but no one cares.


Btw. implementation of the mouse passthrough API in wayland looks like [this](https://github.com/godotengine/godot/blob/0fdbf050e0c7fc7e0a9d42c2a41ee3bfdffbd8f1/platform/linuxbsd/wayland/display_server_wayland.cpp#L1021-L1032), so it's not gonna be any worse than it is.
image

DziubanMaciej:
Anyway, I think it's better to maintain additional 50 lines of code (even multiplied by 3-4 backbends) and have a more convenient GUI, but you do you, it's your project :D You can close the PR if you want. I'm gonna be using this locally.

AsherGlick:
ugg, why does wayland have to do things its own way every time
Ok so this means that the existing solution wont work on wayland either. cool beans.

Alright, if you put this behind a setting so that users can disable if if they are not using x11 then I think it is fine. But I insist that the old pattern stick around for the edge cases of someone trying to use this on mac or windows today because I know of at least 1 user in each of those categories.

DziubanMaciej:
Haha, yeah wayland is a special child :) Okay, I'm glad we came to a conclusion. I added fallbacks to the current approach if either:
- new approach is disabled with a setting (it's disabled by default)
- we're not on Linux+X11
- it failed

DziubanMaciej:
Hey @AsherGlick. Just curious, do you care, or you don't want this change? I've been using it locally and I think it's really cool.

AsherGlick:
@DziubanMaciej I plan to merge the feature as designed. Life is just getting in the way of actually reviewing the code and doing any of the work.

DziubanMaciej:
Alright, no worries :D Thanks.

